PR #10, opened 2017-08-08
Rename modules

ddcprg:


ddcprg:
We need to move the passphrase config of the RPM module to a profile. I'll do this in a separate ticket

PR #13, opened 2017-08-15
Fix broken tunneling

### LINKED ISSUE #12: Tunneling mechanism in Waggle Dance not working
A bug in the passing of HiveConf's meant that the tunnel was being misconfigured

courtsVII:
this commit fixes #12


ddcprg:
Can you add a test for this class?

ddcprg:
Fixes #12

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
```diff
@@ -15,100 +15,45 @@
  */
 package com.hotels.bdp.waggledance.client;
```
courtsVII: Small PR turned into a larger one when Daniel asked for test, this class was doing a lot of things, and the change which introduced the bug couldn't be tested for without decoupling this class

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+      int remotePort) {
+    this.hiveConf = hiveConf;
+    this.tunnelConnectionManager = tunnelConnectionManager;
+    this.sshRoute = sshRoute;
+    this.localHost = localHost;
+    this.remoteHost = remoteHost;
+    this.remotePort = remotePort;
+  }
+
+  public HiveConf getConf() {
```
massdosage: This method isn't just getting a conf object, it's also creating a tunnel so the name is misleading. The name of the class similarly so. I'd expect a "XConfig" class to only hold config and then some other class would create a Tunnel (like a tunnel factory or some such and we already have a few factories here). Perhaps we need a whiteboard session to go over this proposed refactor?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(TunnelConfiguration.class);
+
+  private final HiveConf hiveConf;
+  private final TunnelConnectionManager tunnelConnectionManager;
+  private final String sshRoute;
+  private final String localHost;
+  private final String remoteHost;
+  private final int remotePort;
+
+  public TunnelConfiguration(
```
ddcprg: We don't need this class. The code below can be a method in the builder

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetastoreClientBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+    this.localHost = localHost;
+    return this;
+  }
+
+  public TunnelingMetastoreClientBuilder setRemotePort(Integer remotePort) {
+    this.remotePort = remotePort;
+    return this;
+  }
+
+  public HiveConf getHiveConf() {
```
ddcprg: all these getters are not needed

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetastoreClientBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+    TunnelConfiguration tunnelConfiguration = new TunnelConfiguration(hiveConf, tunnelConnectionManager, sshRoute, localHost,
+        remoteHost, remotePort);
+    HiveConf localHiveConf = tunnelConfiguration.getConf();
+    return clientFromLocalHiveConf(tunnelConnectionManager, localHiveConf);
+  }
+
+  private CloseableThriftHiveMetastoreIface clientFromLocalHiveConf(
+      TunnelConnectionManager tunnelConnectionManager,
+      HiveConf localHiveConf) {
+    CloseableThriftHiveMetastoreIface client = new MetaStoreClientFactory().newInstance(localHiveConf, name,
```
ddcprg: remove `new MetaStoreClientFactory()` and add a method to set this in the builder

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetastoreClientBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+        reconnectionRetries);
+    TunnelingMetastoreClientInvocationHandler tunneledHandler = new TunnelingMetastoreClientInvocationHandler(
+        tunnelConnectionManager, client);
+    return (CloseableThriftHiveMetastoreIface) Proxy.newProxyInstance(getClass().getClassLoader(), INTERFACES,
+        tunneledHandler);
+  }
+
+  public TunnelingMetastoreClientBuilder setHiveConf(HiveConf hiveConf) {
```
ddcprg: We don't use `with` or `set` in Builders. Rename all these from `setAbc` to `abc`

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetastoreClientBuilderTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    //Check that Client is being created from a hiveConf different to that passed into the builder
+    PowerMockito.verifyPrivate(tunnelingMetastoreClientBuilder, times(0))
+                .invoke("clientFromLocalHiveConf", tunnelConnectionManager, hiveConf);
+    PowerMockito.verifyPrivate(tunnelingMetastoreClientBuilder, times(1))
+                .invoke("clientFromLocalHiveConf", eq(tunnelConnectionManager), any(HiveConf.class));
+  }
+
+  @Test
+  public void setHiveConf() throws Exception {
```
ddcprg: we don't want to tests these methods but we do want to test the final result of `build()`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelConfiguration.java
ddcprg: I proposed this to be a method. OK to discuss

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelHandler.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.hotels.bdp.waggledance.api.WaggleDanceException;
+
+public class TunnelHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TunnelHandler.class);
+
+  private TunnelHandler() {}
+
+  public static HiveConf openTunnel(
```
patduin: The method does two things, opening tunnel and creating a new HiveConf as a sideeffect. It should just do one thing. All the hiveConf stuff should be in a separate method.

patduin: why static this makes it hard to test (need for powermock and stuff) a normal class can be injected and mocked in tests

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    PowerMockito.doReturn(tunnelConnectionManager)
+                .when(tunnelingMetaStoreClientFactory, "createTunnelConnectionManager");
+    PowerMockito.doReturn(client)
+                .when(tunnelingMetaStoreClientFactory, "createTunnelingMetastoreClient", eq(localHiveConf), anyString(),
+                    anyInt());
+
+    tunnelingMetaStoreClientFactory.newInstance(hiveConf, "test", 10);
+
+    PowerMockito.verifyPrivate(tunnelingMetaStoreClientFactory, times(1)).invoke("createTunnelConnectionManager");
```
patduin: this makes for very brittle tests can we not verify anything on the instance that is returned?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,80 @@
...
+  private @Spy HiveConf localHiveConf = new HiveConf();
+
+  @Test
+  public void newInstanceWithTunneling() throws Exception {
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_ROUTE.varname, "user@ec2-12-345-678-91.compute-1.amazonaws.com");
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_PRIVATE_KEYS.varname, "private_key");
+    hiveConf.setVar(HiveConf.ConfVars.METASTOREURIS,
+        "thrift://internal-test-shared-hive-metastore-elb-1234567891.us-west-1.elb.amazonaws.com:1234");
+
+    PowerMockito.mockStatic(TunnelHandler.class);
```
patduin: TunnelHandler should be a constructor argument that is mocked in this test. Then you can verify interactions with the tunnelhandler better

courtsVII: Lol, I did this originally then was told to change it to a static method, I'm going to leave it for now, anyone else can chip in their thoughts until their is consensus (all the small changes are slowing down this review)

patduin: I would not use PowerMock unless it is the only way to test something. e.g. legacy code, third party stuff. Leads to very tightly coupled code and brittle tests. I understand this tunneling stuff is hard to test, just saying use with care. But sure a bad test is still better than no test :).

This is your PR, you should take our comments into account and decide for yourself. This is not a massive thing and I don't want to go back and forth so I'll approve and leave it to you.

BTW: I tried to find the comment you refer to or the code change but I can't really find it so not sure what the original code was that triggered you to change it to this, it all seems to have changed quite a bit since then anyway.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
```diff
@@ -15,32 +15,30 @@
...
-import com.jcraft.jsch.JSchException;
 import com.pastdev.jsch.tunnel.TunnelConnectionManager;
 
-import com.hotels.bdp.waggledance.api.WaggleDanceException;
-
 public class TunnelingMetaStoreClientFactory extends MetaStoreClientFactory {
   private static final Logger LOG = LoggerFactory.getLogger(TunnelingMetaStoreClientFactory.class);
 
-  private static class TunnelingMetastoreClientInvocationHandler implements InvocationHandler {
+  private class TunnelingMetastoreClientInvocationHandler implements InvocationHandler {
```
ddcprg: please put `static` back

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactoryTest.java
ddcprg: I don't know why it's been changed but you the problem here is the class that connects to the SSH. If you manage to provide a mock instance of `TunnelConnectionManager` this static method is not needed

patduin: nah mock TunnelHandler I meant

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
patduin: definitely did not mean this :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactoryTest.java
courtsVII: How can you inject TunnelConnectionManagerFactory? The interface doesn't allow it to be injected and the constructor for TunnelConnectionManagerFactory requires the relevant HiveConf passed into newInstance? TunnelHandler can become a concrete class, but we will still have to use power mock to write a test that even captures the bug that this PR fixes, so if thats the only reason to make it not static then I don't really see the point. Power mock was a last resort to work around how this class is tied to the MetaStoreClientFactory interface, I can see why this class didn't have tests originally after working with it.

massdosage:
Did a three way programming session with @ddcprg and @courtsVII and hopefully resolved the final comments from the above PRs. This is hopefully ready to be merged now - please make sure you use the "squash and merge" functionality when merging!

PR #19, opened 2017-08-17
Issue 18

### LINKED ISSUE #18: Update hive dependency to 2.1.1
To match the metastore we're talking to

patduin:
Fixes #18  Update to hive-2.1.1 (~~needs a BeeJu release, so on SNAPSHOT for now~~ Updated to beeju-1.1.0).


waggle-dance-core/pom.xml
```diff
@@ -285,6 +286,16 @@
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot-maven-plugin.version}</version>
         <configuration>
+          <requiresUnpack>
+            <!-- We are hitting an issue introduced in hive-common-2.1.1 and fixed in https://issues.apache.org/jira/browse/HIVE-17155
```
massdosage: Would we want to undo this when we move to a newer version of Hive? If so maybe a TODO with > version X remove?

patduin: It won't break anything if we use Hive-3.0.0 so really it's up to the reality of the future. I would say the comment implies (by fixed) that it can be removed :)

patduin:
fixes #18

waggle-dance-core/pom.xml
massdosage: Ideally we'd remove this later as it's extra code here which we shouldn't need. So I think it wouldn't hurt saying that but fine.

waggle-dance-core/pom.xml
```diff
@@ -253,7 +254,7 @@
     <dependency>
       <groupId>com.hotels</groupId>
       <artifactId>beeju</artifactId>
-      <version>1.0.0</version>
+      <version>${beeju.version}</version>
```
ddcprg: can you move BeeJU to the dependency management section in the parent POM and remove the version from this and the other module?

patduin: sure

PR #21, opened 2017-09-21
Upgrade to Hive 2.3.0

### LINKED ISSUE #20: Upgrade to Hive 2.3.0
Upgrade Hive to version 2.3.0 to make it compatible with AWS EMR 5.8.0

ddcprg:
Fixes #20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## Unreleased
+## [2.0.4]
 ### Changed
+* Upgrade Hive from 2.1.0 to 2.3.0.
 * Fixed bug where tunnel configuration wasn't being applied.
 * Hive dependency updated to 2.1.1 (Needed corresponding BeeJU dependency update as well)
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [2.0.4]
+## [2.1.0]
 ### Changed
 * Upgrade Hive from 2.1.0 to 2.3.0.
 * Depend on latest parent with test.arguments build parameter
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
@@ -1463,4 +1463,36 @@ class FederatedHMSHandler extends FacebookBase implements CloseableIHMSHandler {
     getPrimaryClient().update_master_key(seq_number, key);
   }
 
+  // Hive 2.3.0 methods
+  @Override
+  public List<String> get_tables_by_type(String db_name, String pattern, String tableType)
+    throws MetaException, TException {
+    DatabaseMapping mapping = databaseMappingService.databaseMapping(db_name);
+    return mapping.getClient().get_tables_by_type(mapping.transformInboundDatabaseName(db_name), pattern, tableType);
+  }
+
+  @Override
+  public GetTableResult get_table_req(GetTableRequest req) throws MetaException, NoSuchObjectException, TException {
+    DatabaseMapping mapping = databaseMappingService.databaseMapping(req.getDbName());
+    GetTableResult result = mapping.getClient().get_table_req(mapping.transformInboundGetTableRequest(req));
+    return mapping.transformOutboundGetTableResult(result);
+  }
+
+  @Override
+  public GetTablesResult get_table_objects_by_name_req(GetTablesRequest req)
+    throws MetaException, InvalidOperationException, UnknownDBException, TException {
+    DatabaseMapping mapping = databaseMappingService.databaseMapping(req.getDbName());
+    GetTablesResult result = mapping
+        .getClient()
+        .get_table_objects_by_name_req(mapping.transformInboundGetTablesRequest(req));
+    return mapping.transformOutboundGetTablesResult(result);
+  }
+
+  @Override
+  public CompactionResponse compact2(CompactionRequest rqst) throws TException {
+    DatabaseMapping mapping = databaseMappingService.primaryDatabaseMapping();
+    mapping.checkWritePermissions(rqst.getDbname());
+    return mapping.getClient().compact2(mapping.transformInboundCompactionRequest(rqst));
+  }
+
 }
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
@@ -3,7 +3,7 @@
  *
  * This code is based on Hive's HiveMetaStore:
  *
- * https://github.com/apache/hive/blob/rel/release-2.1.0/metastore/src/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
+ * https://github.com/apache/hive/blob/rel/release-2.3.0/metastore/src/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
  *
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
@@ -23,6 +23,8 @@
  */
 package com.hotels.bdp.waggledance.server;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Condition;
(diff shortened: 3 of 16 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
```diff
@@ -153,8 +156,30 @@ private void startWaggleDance(
       int maxWorkerThreads = hiveConf.getIntVar(ConfVars.METASTORESERVERMAXTHREADS);
       boolean tcpKeepAlive = hiveConf.getBoolVar(ConfVars.METASTORE_TCP_KEEP_ALIVE);
       boolean useFramedTransport = hiveConf.getBoolVar(ConfVars.METASTORE_USE_THRIFT_FRAMED_TRANSPORT);
+      boolean useSSL = hiveConf.getBoolVar(ConfVars.HIVE_METASTORE_USE_SSL);
```
massdosage: Is SSL for the metastore something new they added in 2.3.0?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -1463,4 +1463,36 @@ public void update_master_key(int seq_number, String key) throws NoSuchObjectExc
     getPrimaryClient().update_master_key(seq_number, key);
   }
```
massdosage: It looks like they've just added some new methods so this should be backwards compatible?

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -277,13 +277,13 @@ public void typicalWithGraphite() throws Exception {
     assertMetric(metrics,
         "graphitePrefix.counter.com.hotels.bdp.waggledance.server.FederatedHMSHandler.get_all_databases.all.success.count 2");
     assertMetric(metrics,
-        "graphitePrefix.counter.com.hotels.bdp.waggledance.server.FederatedHMSHandler.get_table.primary.calls.count 1");
+        "graphitePrefix.counter.com.hotels.bdp.waggledance.server.FederatedHMSHandler.get_table_req.primary.calls.count 1");
```
massdosage: Hmmm, so does this mean that getTable() at the client level is being translated into something new (get_table_req) instead of get_table at the Thrift layer? What happened to the get_table code path? How different are they? Is this the part we're worried might not be backwards compatible?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
ddcprg: this code is new, I suppose it was added recently

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
ddcprg: I guess if an 2.1 client fetches a table then get_table will be invoked. It should be fine as long as both the actual metastore and WD use the same Hive version. The problem is if the real metastore runs 2.1, WD 2.3 and the client 2.3. I'll be running functional tests next week so we can decided wether we start maintaining a development branch for each of these versions and possibly future versions

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
ddcprg: as per my previous comment

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## Unreleased
+## [2.0.4]
```
massdosage: I would go for a minor version bump for a change like this as Hive is a big part of what WD is all about - i.e. 2.1.0.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
massdosage: OK, so what we're doing here is enabling the client to talk to us using SSL and we then in turn talk to the proxied HMS using SSL too?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
```diff
@@ -153,8 +156,30 @@ private void startWaggleDance(
...
 
-      TServerTransport serverTransport = tcpKeepAlive ? new TServerSocketKeepAlive(waggleDanceConfiguration.getPort())
+      TServerSocket serverSocket = null;
+
+      // enable SSL support for HMS
+      List<String> sslVersionBlacklist = new ArrayList<>();
```
massdosage: Is there any way at least some of this code could be factored into a separate private method? This method is getting quite long now...

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
```diff
@@ -153,8 +156,30 @@ private void startWaggleDance(
...
+      for (String sslVersion : hiveConf.getVar(ConfVars.HIVE_SSL_PROTOCOL_BLACKLIST).split(",")) {
+        sslVersionBlacklist.add(sslVersion);
+      }
+      if (!useSSL) {
```
massdosage: Like maybe lines 168-180 could go into a method which returns the correct type of server socket?

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
massdosage: OK, we discussed this offline, looks like if things are running the same version it's all fine and the backwards compatible use cases that we care about work too so this is fine.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
massdosage: Also, any way we can test any of this? (I understand it might be quite hard...)

CHANGELOG.md
ddcprg: OK

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
ddcprg: If we start tidying up the code I'll be harder to compare with the original class in future releases. I prefer to leave the code as is so we can pick copy easy copy bits of code

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
ddcprg: I see whether is possible to test this somehow

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
massdosage: OK, that makes sense, can you then add a comment above this code snippet saying which class in Hive this was copied from so it's easier to line them up in future?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
massdosage: You mean the Hive project doesn't have a nice clean test for this that we can copy? ;P

ddcprg: strangely not

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
ddcprg: I've moved this to its own method as there are other pieces that were moved already

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
ddcprg: done

CHANGELOG.md
massdosage: So are you going to bump it? Here and in the poms.

ddcprg: yes

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
```diff
@@ -187,6 +195,28 @@ private void startWaggleDance(
     LOG.info("Waggle Dance has stopped");
   }
 
+  private TServerSocket createServerSocket(boolean useSSL, int port) throws IOException, TTransportException {
```
massdosage: Why not remove useSSL as a param (we generally try avoid boolean params if possible) and retrieve the value from the conf here? You're already doing it for some other stuff in this method and you're not using the useSSL variable outside of it.

ddcprg: done

PR #26, opened 2017-10-02
Default GC setting update

patduin:
changed gc settings, idea is to cope with bigger requests better

Lowered overal heap size to give more to OS and set XX:G1ReservePercent (default is 10), which I understand helps in preserving a bigger chunk before GC is called.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## [2.1.0] TBD
+## [2.1.0] 2017-10-02
 ### Changed
+* Changed the default GC settings, tuning for our environment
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## [2.1.0] TBD
+## [2.1.0] 2017-10-02
 ### Changed
+* Changed the default GC settings, tuning for our environment
```
massdosage: "our" environment doesn't mean much to anyone else, rather just describe what the defaults have changed to (less overall ram, more reserved percent) and/or what this change to the default settings gets a user

patduin: good point, changed it

PR #27, opened 2017-10-03
Improve memory usage

### LINKED ISSUE #24: Improve memory usage in federated metastore request/response forwards
**_Spike_**

**Overview**
At the moment the database mappings create copies of the Thrift objects it processes. Some of these objects can be huge and the creation of copy will demand memory for the request. As we only need the original object to determine which megastore will process the request we could mutate the current object instead of making a copy.

**Acceptance criteria**
Evaluate memory usage before and after the change.

ddcprg:
Fixes #24 

After monitoring both the current version and this version we noticed an improvement in the memory usage reported WD.

I also took some time to add few missing bits like the method annotations in the handler and additional tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 ## [2.1.0]
 ### Changed
 * Upgrade Hive from 2.1.0 to 2.3.0.
-* Depend on latest parent with test.arguments build parameter
+* Depend on latest parent with test.arguments build parameter.
 * Fixed bug where tunnel configuration wasn't being applied.
-* Hive dependency updated to 2.1.1 (Needed corresponding BeeJU dependency update as well)
+* Hive dependency updated to 2.1.1 (Needed corresponding BeeJU dependency update as well).
+* `DatabaseMapping`s no longer make copies of the Thrift objects, it mutates the original objects instead.
 ### Removed
 * Removed SessionFactorySupplierFactory.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,13 @@
-## [2.1.1]
+## [2.2.0] Unreleased
 ### Changed
 * `DatabaseMapping`s no longer make copies of the Thrift objects, it mutates the original objects instead.
-* Upgrade Spring, BeeJU and other dependencies
+* Upgrade Spring, BeeJU and other dependencies.
 
 ## [2.1.0] 2017-10-03
 ### Changed
 * Changed the default GC settings, less heap, more reserved percentage, works better with large requests.
 * Upgrade Hive from 2.1.0 to 2.3.0.
-* Depend on latest parent with test.arguments build parameter.
+* Depend on latest parent with `test.arguments` build parameter.
 * Fixed bug where tunnel configuration wasn't being applied.
 * Hive dependency updated to 2.1.1 (Needed corresponding BeeJU dependency update as well).
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [2.1.1]
 ### Changed
 * `DatabaseMapping`s no longer make copies of the Thrift objects, it mutates the original objects instead.
-* Upgrade BeeJU
+* Upgrade Spring, BeeJU and other dependencies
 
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
...
+### Changed
+* `DatabaseMapping`s no longer make copies of the Thrift objects, it mutates the original objects instead.
+* Upgrade Spring, BeeJU and other dependencies
+
 ## [2.1.0] 2017-10-03
 ### Changed
 * Changed the default GC settings, less heap, more reserved percentage, works better with large requests.
 * Upgrade Hive from 2.1.0 to 2.3.0.
-* Depend on latest parent with test.arguments build parameter
+* Depend on latest parent with test.arguments build parameter.
```
massdosage: Nice, well spotted on the missing full stops! ;) However I have now noticed that you are missing one on line 4!

pom.xml
```diff
@@ -35,14 +35,14 @@
     <cobertura.version>2.0.3</cobertura.version>
     <cobertura.maven.plugin.version>2.6</cobertura.maven.plugin.version>
     <maven.release.plugin.version>2.5.3</maven.release.plugin.version>
-    <spring-platform.version>2.0.8.RELEASE</spring-platform.version>
+    <spring-platform.version>Brussels-SR5</spring-platform.version>
```
massdosage: Really? They've moved from semantic versioning to names? Or is this some special pre-release versioning scheme they've come up with?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMapping.java
```diff
@@ -101,7 +107,7 @@
 
   DropPartitionsRequest transformInboundDropPartitionRequest(DropPartitionsRequest req);
 
-  DropPartitionsResult transforOutboundDropPartitionsResult(DropPartitionsResult dropPartitionsResult);
+  DropPartitionsResult transformOutboundDropPartitionsResult(DropPartitionsResult dropPartitionsResult);
```
massdosage: Good catch on the typo, the original reviewers should be punished!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
```diff
@@ -3,7 +3,7 @@
  *
  * This code is based on Hive's HiveMetaStore:
  *
- * https://github.com/apache/hive/blob/rel/release-2.3.0/metastore/src/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
+ * https://github.com/apache/hive/blob/rel/release-2.1.0/metastore/src/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
```
massdosage: Was this intentional to go back to 2.1.0 instead of 2.3.0?

CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
+## [2.1.1]
```
massdosage: Lets put "Unreleased" in place of the date so it's clearer if this goes into master but we don't have time to do a release. Also, I think the changes are big enough that we go for 2.2.0.

pom.xml
massdosage: Went and had a look at their site, so yes, they seem to have moved to "names" for these "meta" releases of their platform. I prefer SemVer but hey, nothing we can do about.

CHANGELOG.md
ddcprg: OK

CHANGELOG.md
ddcprg: I'll add it

pom.xml
ddcprg: Nothing we can do

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
ddcprg: Merge issue probably?

PR #30, opened 2017-10-30
issue-29: fixed missing STATUS from rest endpoint

patduin:


waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
```diff
@@ -38,7 +39,7 @@
   private @NotBlank String remoteMetaStoreUris;
   private @Valid MetastoreTunnel metastoreTunnel;
   private @NotNull AccessControlType accessControlType = AccessControlType.READ_ONLY;
-  private transient @NotNull MetaStoreStatus status = MetaStoreStatus.UNKNOWN;
+  private transient @JsonProperty @NotNull MetaStoreStatus status = MetaStoreStatus.UNKNOWN;
```
patduin: The actual fix, the rest is test related. Explicitly setting it as a JsonProperty. Following: https://github.com/FasterXML/jackson-databind/issues/1184

waggle-dance-rest/pom.xml
```diff
@@ -80,6 +80,12 @@
       <groupId>org.skyscreamer</groupId>
       <artifactId>jsonassert</artifactId>
       <scope>test</scope>
+      <exclusions>
```
massdosage: Do you need to do this in order to prevent a conflicting version of Jackson being pulled in?

patduin: The rest controller test prints a warning about two similar class in two different jars on the classpath. It Adviced to remove one for deterministic test sake :). I don't need to do this for the fix.

PR #37, opened 2017-11-24
Issue 34

patduin:
Pooling client connections.


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
```diff
@@ -0,0 +1,119 @@
...
+import com.hotels.bdp.waggledance.api.model.AbstractMetaStore;
+import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIface;
+
+/**
+ * Simple delegate to get rid of generics and add unchecked {@link #borrowObjectUnchecked(AbstractMetaStore)},
+ * {@link #returnObjectUnchecked(AbstractMetaStore, CloseableThriftHiveMetastoreIface) methods.
+ */
+public class MetaStoreClientPool implements KeyedObjectPool<AbstractMetaStore, CloseableThriftHiveMetastoreIface> {
+
+  private final static Logger log = LoggerFactory.getLogger(MetaStoreClientPool.class);
```
courtsVII: log -> LOG in keeping with the rest of the code

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
```diff
@@ -78,11 +84,18 @@ public String getDatabasePrefix() {
 
   @Override
   public void close() throws IOException {
-    client.close();
```
courtsVII: Does this still need to be closed explicitly at some point?

patduin: Connection will be monitored so it will be closed, but probably best to do it in the pool and poolFactory explicitly as well. Good practice

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
patduin: heh now my eclipse macro has the wrong default :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
```diff
@@ -31,7 +31,7 @@
  */
 public class MetaStoreClientPool implements KeyedObjectPool<AbstractMetaStore, CloseableThriftHiveMetastoreIface> {
 
-  private final static Logger log = LoggerFactory.getLogger(MetaStoreClientPool.class);
+  private final static Logger LOG = LoggerFactory.getLogger(MetaStoreClientPool.class);
```
massdosage: OMG, going back the other way now! ;P

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
```diff
@@ -0,0 +1,120 @@
...
+    pool.close();
+  }
+
+  public void returnObjectUnchecked(AbstractMetaStore metaStore, CloseableThriftHiveMetastoreIface client) {
+    LOG.debug("returning client to pool for key: {}, activeCount (total): {} ({}), idleCount (total): {} ({})",
+        metaStore.getName(), getNumActive(metaStore), getNumActive(), getNumIdle(metaStore), getNumIdle());
+    try {
+      pool.returnObject(metaStore, client);
+    } catch (Exception e) {
+      LOG.warn("Could not return object to the pool, metastore is '" + metaStore.getName() + "'", e);
```
massdosage: I assume it's on purpose that you don't rethrow the error here?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
```diff
@@ -53,6 +56,9 @@ public String transformOutboundDatabaseName(String databaseName) {
 
   @Override
   public ThriftHiveMetastore.Iface getClient() {
```
massdosage: Are calls to this single threaded?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
```diff
@@ -78,11 +84,18 @@ public String getDatabasePrefix() {
 
   @Override
   public void close() throws IOException {
-    client.close();
+    if (client != null) {
```
massdosage: Same here (regarding single threaded)

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -524,4 +527,25 @@ public void doesNotOverwriteConfigOnShutdownAfterAddingFederation() throws Excep
     assertThat(remoteMetastore.getMappedDatabases().get(0), is(REMOTE_DATABASE));
   }
 
+  @Test
+  public void restAPIGetStatus() throws Exception {
```
massdosage: restApiGetStatus

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -524,4 +527,25 @@ public void doesNotOverwriteConfigOnShutdownAfterAddingFederation() throws Excep
...
+
+    RestTemplate rest = new RestTemplateBuilder().build();
+    PrimaryMetaStore primaryMetastore = rest.getForObject("http://localhost:18000/api/admin/federations/primary",
+        PrimaryMetaStore.class);
+    assertThat(primaryMetastore.getStatus(), is(MetaStoreStatus.AVAILABLE));
+    FederatedMetaStore federatedMetastore = rest
+        .getForObject("http://localhost:18000/api/admin/federations/waggle_remote", FederatedMetaStore.class);
+    assertThat(federatedMetastore.getStatus(), is(MetaStoreStatus.AVAILABLE));
+  }
+
```
massdosage: Is it worth having a test for UNAVAILABLE?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
patduin: Don't MIX and match I guess, I gave up :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
patduin: yes, nothing has changed here. Request comes in a MetastoreMapping is being made and the request is handled by the client, nothing asynchronous going on here.

On an aside. This is bad method that's leaking the client. Not trivial to get rid of it though.

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
patduin: Not here, golden path integeration test. The rest module has tests for the service for more specific stuff. This is just here to make sure all the stuff is wired up correctly.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
patduin: yeah should be fine

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
patduin: yes I don't really care. This would be something very weird e.g. the key is incorrect. As long as I can keep getting clients from the pool I'm fine

patduin:
@teabot yes they maintain state but that means they are not threadsafe. A pool does nothing more but give a connect to a single thread. A client will still only be used by a single thread. So it is the same as doing sequential requests to the same client.
A pool fixes the expensive cost of making a client. It made sense while trying to reuse the client factory in the rest service.
I'm gonna do a quick check to see how expensive a tunneled client is to make. If it seems fine I'll refactor the pool out. I've refactored some other stuff in here that I kind of like and got some other changes incoming as well while I was testing concurrent access.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/pool/MetaStoreClientPool.java
```diff
@@ -0,0 +1,123 @@
...
+
+import com.hotels.bdp.waggledance.api.WaggleDanceException;
+import com.hotels.bdp.waggledance.api.model.AbstractMetaStore;
+import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIface;
+
+/**
+ * Simple delegate to get rid of generics and add unchecked {@link #borrowObjectUnchecked(AbstractMetaStore)},
+ * {@link #returnObjectUnchecked(AbstractMetaStore, CloseableThriftHiveMetastoreIface) methods.
+ */
+public class MetaStoreClientPool implements KeyedObjectPool<AbstractMetaStore, CloseableThriftHiveMetastoreIface> {
```
ddcprg: did you check this project? -> https://github.com/aloha-app/thrift-client-pool-java
I'm not sure if it could be used instead of implementing our own pool

patduin: I had a quick look, but went for commons pool implementation. It all ends up being the same tbh, you need to create some factory for the clients and then configure you're pool accordingly.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/HMSHandlerProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+  public HMSHandlerProcessor(IHMSHandler handler, TProcessor delegate) throws SecurityException, NoSuchFieldException,
+      IllegalArgumentException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
+    this.handler = handler;
+    this.delegate = delegate;
+  }
+
+  @Override
+  public boolean process(final TProtocol in, final TProtocol out) throws TException {
+    try {
+      return delegate.process(in, out);
```
patduin: ok so this is a bad idea. Multiple process call might be done through one client. Cannot close the handler here.

Issue #34, opened 2017-11-14
Status health checks doesn't support tunnels


patduin:
_As a user of Waggle Dance
I'd like to see the status of a metastore that is connected to via a tunnel_

Acceptance Criteria:
* Fix the STATUS check returned by the rest api to support a tunnelled connection.

From looking at the code I think it would be great (if possible) if we can reuse the connections already setup in the MetastoreMappingImpl. That way the status check is a more accurate representation of available connections.

patduin:
I've created a branch for this issue already that added some logging and a quick fix to unblock us. 
Branch is here : https://github.com/HotelsDotCom/waggle-dance/tree/issue-34

PR #38, opened 2017-11-24
Reuse client factory for rest and handler usage

patduin:
Alternative to https://github.com/HotelsDotCom/waggle-dance/pull/37 if we don't like the pool.

Update:
Leaving the pool.
Going for this solution now. Fixes issue #34
Please review and I'll get on a release.

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/SimpleFederationStatusService.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/SimpleFederationStatusService.java
@@ -15,21 +15,23 @@
  */
 package com.hotels.bdp.waggledance.mapping.service.impl;
 
-import org.apache.hadoop.hive.conf.HiveConf;
-import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
-import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import com.hotels.bdp.waggledance.api.federation.service.FederationStatusService;
+import com.hotels.bdp.waggledance.api.model.AbstractMetaStore;
 import com.hotels.bdp.waggledance.api.model.MetaStoreStatus;
+import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIface;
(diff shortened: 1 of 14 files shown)


patduin:
@massdosage yeah there's still quite a bit of refactoring to be done having slept on it they won't be that far off tbh. But yes could do it in two steps.

ddcprg:
I've not fully looked at the other PR but this looks neater. I'd would also go for this change first and then we can see how we can make this more efficiently.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/SimpleFederationStatusService.java
```diff
@@ -44,17 +46,13 @@ private HiveConf newHiveConf(String metaStoreUris) {
...
-  public MetaStoreStatus checkStatus(String metaStoreUris) {
-    HiveConf hiveConf = newHiveConf(metaStoreUris);
-    HiveMetaStoreClient client = null;
-    try {
-      client = new HiveMetaStoreClient(hiveConf);
+  public MetaStoreStatus checkStatus(AbstractMetaStore abstractMetaStore) {
+    try (CloseableThriftHiveMetastoreIface client = metaStoreClientFactory.newInstance(abstractMetaStore)) {
+      if (!client.isOpen()) {
+        return MetaStoreStatus.UNAVAILABLE;
+      }
```
massdosage: Where does the closing of the client happen now?

waggle-dance-rpm/pom.xml
```diff
@@ -26,7 +26,7 @@
         <artifactId>maven-dependency-plugin</artifactId>
         <executions>
           <execution>
-            <id>circus-train-bin</id>
+            <id>waggle-dance-bin</id>
```
massdosage: Heh, good spot.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/SimpleFederationStatusService.java
patduin: Try with resources, java does it for me \o/

massdosage: Aah, of course!

PR #39, opened 2017-12-14
Doclint and travis profiles now in parent

massdosage:
I'll replace SNAPSHOT version of parent when HotelsDotCom/hotels-oss-parent#10 is merged and released

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.hotels</groupId>
     <artifactId>hotels-oss-parent</artifactId>
-    <version>2.0.6-SNAPSHOT</version>
+    <version>2.0.6</version>
   </parent>
 
   <artifactId>waggle-dance-parent</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -193,35 +193,4 @@
...
-          <plugin>
-            <groupId>org.eluder.coveralls</groupId>
-            <artifactId>coveralls-maven-plugin</artifactId>
-            <version>4.3.0</version>
-          </plugin>
-        </plugins>
-      </build>
-    </profile>
-    <profile>
-      <id>doclint-java8-disable</id>
```
ddcprg: are we not using this?

massdosage: We are, it's already in the parent so no need to have it here again.

ddcprg: ok, I see

ddcprg:
Add a line in CHANGELOG.md

massdosage:
Since the change isn't user facing or affect waggle-dance itself in any way I didn't think it was worth adding to the changelog.

PR #41, opened 2018-02-09
Clarified where the rpm is coming from

patduin:


README.md
```diff
@@ -77,7 +77,7 @@ Log messages will be output to the standard output.
 
 It is intended that Waggle Dance is run as a service serving as a proxy to different metastores. The primary configured metastore is the only one to which you may also write data via Hive.
 
-The RPM package needs to be downloaded and is available in Maven Central the file can be installed using `yum`.
+The RPM package needs to be downloaded and is available in [![Maven Central RPM](https://maven-badges.herokuapp.com/maven-central/com.hotels/waggle-dance-rpm/badge.svg?subject=com.hotels:waggle-dance.rpm)](https://maven-badges.herokuapp.com/maven-central/com.hotels/waggle-dance-rpm) (it is not available in any rpm repo) the file can be installed using `yum`.
```
massdosage: Mmm, that Maven link is a bit confusing if you're not very familiar with Maven. It shows the XML fragment for adding a dependency on it. Ideally we would link straight to http://repo1.maven.org/maven2/com/hotels/waggle-dance-rpm/2.2.2/waggle-dance-rpm-2.2.2.rpm (and same above with the .tgz) but I don't know how we can do that without harcoding the numbers in this file. Saying "it is not available in any RPM repo" is also a bit confusing, what we're really trying to say here is that "it's not available in any of the RPM-based Linux distribution's repos" but that's a bit of a mouthful. I think here we should just say something along the lines of "an RPM version of WaggleDance is available for download [here]. Once downloaded this can be installed using your distribution's packaging tool, e.g. `yum`:"

patduin: done

PR #50, opened 2018-03-07
SSH timeout added

### LINKED ISSUE #49: Allow users to configure a session timeout for SSH tunnels
The admin endpoint hangs when the SSH tunnels uses an unreachable host.

ddcprg:
Fixes #49 

I'll be running functional tests today, please take a look in the meantime

### DIFF
--- a/waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnel.java
+++ b/waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnel.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,12 +26,14 @@ public class MetastoreTunnel {
 
   private static final int DEFAULT_PORT = 22;
   private static final String DEFAULT_LOCALHOST = "localhost";
+  private static final int DEFAULT_TIMEOUT_MILLIS = 60000; // 1 minute
 
   private @NotBlank @TunnelRoute String route;
   private @Min(1) @Max(65535) int port = DEFAULT_PORT;
   private String localhost = DEFAULT_LOCALHOST;
   private @NotBlank String privateKeys;
   private @NotBlank String knownHosts;
+  private @Min(0) int timeout = DEFAULT_TIMEOUT_MILLIS;
 
   public String getRoute() {
     return route;
@@ -71,6 +73,14 @@ public class MetastoreTunnel {
 
   public void setKnownHosts(String knownHosts) {
     this.knownHosts = knownHosts;
+  }
+
+  public int getTimeout() {
+    return timeout;
+  }
+
+  public void setTimeout(int timeout) {
+    this.timeout = timeout;
   }
 
 }
--- a/waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
+++ b/waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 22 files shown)


waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnel.java
```diff
@@ -26,12 +26,14 @@
...
   private static final int DEFAULT_PORT = 22;
   private static final String DEFAULT_LOCALHOST = "localhost";
+  private static final int DEFAULT_TIMEOUT_MILLIS = 60000; // 1 minute
 
   private @NotBlank @TunnelRoute String route;
   private @Min(1) @Max(65535) int port = DEFAULT_PORT;
   private String localhost = DEFAULT_LOCALHOST;
   private @NotBlank String privateKeys;
   private @NotBlank String knownHosts;
+  private @Min(0) int timeout = DEFAULT_TIMEOUT_MILLIS;
```
massdosage: I think we should either name this timeoutMillis to be clear what unit it is or we introduce two values here - one which is a TimeUnit and the other being the value.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
```diff
@@ -49,6 +50,15 @@ public void typical() {
     assertThat(violations.size(), is(0));
   }
 
+  @Test
+  public void infiniteTimeout() {
```
massdosage: We should document in the README that setting it to zero means no timeout.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
```diff
@@ -148,4 +158,13 @@ public void blankPrivateKey() {
     assertThat(violations.size(), is(1));
   }
 
+  @Test
+  public void negativeTunnel() {
```
massdosage: negativeTimeout

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SessionFactorySupplier.java
```diff
@@ -31,15 +34,40 @@
...
+  private final int sshTimeout;
   private DefaultSessionFactory sessionFactory = null;
 
+  @Deprecated
   public SessionFactorySupplier(int sshPort, String knownHosts, List<String> identityKeys) {
+    this(sshPort, knownHosts, identityKeys, 0);
+  }
+
+  public SessionFactorySupplier(int sshPort, String knownHosts, List<String> identityKeys, int sshTimeout) {
+    Preconditions.checkArgument(0 <= sshPort && sshPort <= 65535, "Invalid SSH port number " + sshPort);
```
massdosage: Why do we have this check both here and in the TunnelHandler  but we only have the timeout check here and not in the TunnelHandler ?

ddcprg: I'm improving the code, the old version does not check... the constructor is public

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
ddcprg: yeah, typo.. will rename

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
ddcprg: will document

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnel.java
ddcprg: Even if the docs say the units is millis?

massdosage: Yeah, I guess it's fine as long as its clearly documented.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SessionFactorySupplier.java
massdosage: OK

ddcprg:
This just got uglier but it works. Please take a look again and I'll try to do a release tomorrow.

PR #68, opened 2018-04-25
Revert "issue-62: fix queries against views"

courtsVII:
Reverts HotelsDotCom/waggle-dance#66


courtsVII:
Reverted to address @teabot code review

Issue #62, opened 2018-04-10
Select * from <VIEW> on federated metastore does not work.


pradeepbhadani:
If I have a View `View1` over Table `T1` and `T2` under hive database `db1` in federated metastore `FM1`. Following command fails with `Table Not Found Exception`

`Select * from fm1_db1.view1 ;`
This will fail with Table `T1` & `T2` Not Found.

Additional Info:
* WD is running in prefixed mode.

pradeepbhadani:
Looks like WD is not able to add prefix to table `T1` and `T2` when it expands view to table level.

PR #72, opened 2018-05-14
Using view transform logic for all Table functions

patduin:
Fixing issue #62

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright (C) 2016-2018 Expedia Inc.
+ *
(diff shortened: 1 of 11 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
```diff
@@ -0,0 +1,80 @@
...
+
+  INSTANCE;
+
+  @Override
+  public String transformOutboundDatabaseName(MetaStoreMapping metaStoreMapping, String query) {
+    ASTNode root;
+    try {
+      root = ParseUtils.parse(query);
+    } catch (ParseException e) {
+      throw new WaggleDanceException(e.getMessage());
```
massdosage: Is there a reason for not adding the exception here? Generally I like being able to see the stack trace of the cause unless it doesn't add anything (which is rarely the case)

patduin: I agree, I'll throw in the exception

PR #75, opened 2018-05-16
using commons-lang Pair

### LINKED ISSUE #74: Views error: java.lang.NoClassDefFoundError: com/sun/tools/javac/util/Pair
Issue in handling views that are generated via subqueries. WD will throw: java.lang.NoClassDefFoundError: com/sun/tools/javac/util/Pair

Related to #62

patduin:
Fixes #74

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/parse/ASTConverter.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/parse/ASTConverter.java
@@ -30,11 +30,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.hive.ql.parse.ASTNode;
 import org.apache.hadoop.hive.ql.parse.HiveParser;
 import org.apache.http.annotation.Experimental;
-
-import com.sun.tools.javac.util.Pair;
 
 import com.hotels.bdp.waggledance.parse.Rule.RuleSet;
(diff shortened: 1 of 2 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/parse/ASTConverter.java
```diff
@@ -30,12 +30,12 @@
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
```
patduin: this class gets pulled it via hive-exec jar, not great but I don't want to make a too involved change now (We shouldn't use hive-exec but hive-exec-core.jar etc.... out of scope).

massdosage: Sure, using another Pair that is already on the classpath sounds like the quickest fix for now. Later we can decide if we want to introduce a new class or use something else.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/parse/ASTConverter.java
```diff
@@ -30,12 +30,12 @@
...
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.hive.ql.parse.ASTNode;
 import org.apache.hadoop.hive.ql.parse.HiveParser;
 import org.apache.http.annotation.Experimental;
 
-import com.sun.tools.javac.util.Pair;
```
ddcprg: I didn't spot this in the other PR, we should not use classes from Java tools unless we are doing something very specific with the JVM..  this is not part of the language packages

PR #76, opened 2018-05-17
Added client call to isOpen so the whole connection roundtrip is tested

### LINKED ISSUE #73: Waggle Dance fail to respond if one of the metastore becomes unresponsive
If I have one Primary Metasotre and `n` federated metastore and one of the metastore stop responding then client connecting to Waggle Dance thrift endpoint gets `Thrift Exception` and unable to query data/metadata of rest of the metastore.

Additional Info:
* WD is running in prefixed mode.

patduin:
fixes #73

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClient.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClient.java
@@ -155,6 +155,7 @@ class ThriftMetastoreClient implements Closeable {
           client = new ThriftHiveMetastore.Client(protocol);
           try {
             transport.open();
+
             LOG.info("Opened a connection to metastore '"
                 + store
                 + "', total current connections to all metastores: "
@@ -217,7 +218,7 @@ class ThriftMetastoreClient implements Closeable {
     }
     // Transport would have got closed via client.shutdown(), so we don't need this, but
     // just in case, we make this call.
-    if (isOpen()) {
+    if (transport != null && transport.isOpen()) {
       transport.close();
       transport = null;
     }
@@ -225,7 +226,15 @@ class ThriftMetastoreClient implements Closeable {
   }
 
   public boolean isOpen() {
-    return transport != null && transport.isOpen();
+    if (transport != null && transport.isOpen() && client != null) {
+      try {
+        client.getStatus();
+        return true;
+      } catch (TException e) {
+        return false;
+      }
+    }
+    return false;
   }
 
   protected ThriftHiveMetastore.Iface getClient() {
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
@@ -113,7 +113,7 @@ public class TunnelingMetaStoreClientFactory implements MetaStoreClientFactory {
         .get(localHiveConf);
 
     LOG.info("Metastore URI {} is being proxied to {}", localHiveConf.getVar(HiveConf.ConfVars.METASTOREURIS),
-        localHiveConf.getVar(HiveConf.ConfVars.METASTOREURIS));
+        hiveConf.getVar(HiveConf.ConfVars.METASTOREURIS));
 
     HiveMetaStoreClientSupplier supplier = new HiveMetaStoreClientSupplier(localHiveConf, name, reconnectionRetries);
     return (CloseableThriftHiveMetastoreIface) tunnelableFactory.wrap(supplier, METHOD_CHECKER, localHost, localPort,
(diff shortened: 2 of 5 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClient.java
```diff
@@ -217,15 +218,23 @@ public void close() {
...
       transport = null;
     }
     LOG.info("Closed a connection to metastore, current connections: " + CONN_COUNT.decrementAndGet());
   }
 
   public boolean isOpen() {
-    return transport != null && transport.isOpen();
+    if (transport != null && transport.isOpen() && client != null) {
+      try {
+        client.getStatus();
```
patduin: more or less just picked some call that can be made, doesn't really matter this one seems fairly lightweight

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -228,7 +228,9 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
       if (mapping == null) {
         continue;
       }
-      mappings.put(mapping, prefixWithPattern.getValue());
+      if (includeInResults(mapping)) {
```
patduin: noticed that this was missing, so added it and tests accordingly

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClient.java
```diff
@@ -217,15 +218,23 @@ public void close() {
     }
     // Transport would have got closed via client.shutdown(), so we don't need this, but
     // just in case, we make this call.
-    if (isOpen()) {
+    if (transport != null && transport.isOpen()) {
```
ddcprg: the new condition is not needed

ddcprg: why not to leave it as it was?

ddcprg: ok, I see.. leave the other review and approve

patduin: the client gets closed a couple of lines above, the body of the if is just concerned with the transport

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClient.java
```diff
@@ -217,15 +218,23 @@ public void close() {
...
   }
 
   public boolean isOpen() {
-    return transport != null && transport.isOpen();
+    if (transport != null && transport.isOpen() && client != null) {
+      try {
+        client.getStatus();
+        return true;
+      } catch (TException e) {
+        return false;
```
massdosage: Do we want to log something if this happens?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
```diff
@@ -113,7 +113,7 @@ public CloseableThriftHiveMetastoreIface newInstance(HiveConf hiveConf, String n
         .get(localHiveConf);
 
     LOG.info("Metastore URI {} is being proxied to {}", localHiveConf.getVar(HiveConf.ConfVars.METASTOREURIS),
-        localHiveConf.getVar(HiveConf.ConfVars.METASTOREURIS));
+        hiveConf.getVar(HiveConf.ConfVars.METASTOREURIS));
```
massdosage: Oh dear, this looks like it was wrong too, what would this cause to happen?

patduin: it is just a log message it would say, localhost:1234 is being proxied to localhost:1234, but it is fixed now. ;)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClient.java
patduin: good question, I'll have a look if it isn't printed already, it might be due to how the logic is wired but perhaps it needs some cleaning up

patduin: I changed the logic around a bit, but the error gets correctly logged in the logs (for every retry)

patduin:
Something to keep in mind (this PR doesn't address this) the performance degrades when a connections is down. For a request that involves all metastores (like `show databases`, or our healtcheck call) a connection is established and if this fails it is reattempted 2 more times (this is hardcoded) with a 1 second sleep. So you add  latency of 2 seconds per unreachable metastore to a request. 
A request that doesn't involve all metastore obviously doesn't have this.

PR #79, opened 2018-05-22
Use hcommon-hive-metastore

### LINKED ISSUE #78: Extract common Hive Metastore code into a library

courtsVII:
fixes #78


patduin:
Why are we pulling out code that we are not sharing? 
I'm gonna go conservative here and say put code in the common package if it is used or is going to be used in multiple projects otherwise leave it here. I'm not all that convinced the "generic" hive code in WD is particularly useful for other projects. There are certain cons to having the code in a separate library, releasing everything now becomes more difficult, dependencies become a bit more complex.

massdosage:
Yeah, that's a good point @patduin. I think we all got a bit overly enthusiastic here ;) So, yes, the stuff from Circus Train that is needed in order to be used from both Circus Train and Housekeeping is worth pulling out. For Waggle Dance if there are exact duplicate classes (or very similar) then I think it's worth adapting Waggledance to use the new shared library. I guess it can be quite tricky to line them up and then there might be some coupling with additional WaggleDance classes so they ended up being copied over too. It's probably worth doing a step back and just focusing on the Circus Train and Housekeeping side of things and get all that working first, then seeing how WaggleDance could be adapted to use the new shared library.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/metastore/ReconnectingMetaStoreClientFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.apache.hadoop.hive.conf.HiveConf;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.hotels.hcommon.hive.metastore.MetaStoreClientException;
+import com.hotels.hcommon.hive.metastore.client.api.CloseableMetaStoreClient;
+import com.hotels.hcommon.hive.metastore.client.api.MetaStoreClientFactory;
+
+class ReconnectingMetaStoreClientFactory implements MetaStoreClientFactory {
```
massdosage: Where do these "ReconnectingXxx" classes come from? I thought this PR was going to be just modifying a few classes to use new common lib for wherever there was overlap as opposed to adding new classes.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/MetaStoreUriNormaliser.java
```diff
@@ -0,0 +1,40 @@
...
+package com.hotels.bdp.waggledance.util;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Set;
+import java.util.TreeSet;
+
+import com.google.common.base.Joiner;
+
+public class MetaStoreUriNormaliser {
```
massdosage: Similar, why is this new class being introduced?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImplTest.java
```diff
@@ -13,17 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package com.hotels.bdp.waggledance.mapping.model;
 
+import static com.hotels.bdp.waggledance.api.model.AbstractMetaStore.newFederatedInstance;
```
massdosage: import order wrong again

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/MetaStoreUriNormaliser.java
patduin: I see am I not the only confused. I did some digging seems this was a static method before.
It is just hard to follow when stuff is moving and refactored in the same PR. @courtsVII perhaps two PR's one after the other are a better idea. One for moving stuff as is, one for doing the refactoring. Makes it a lot easier to review/trace back later etc...

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/metastore/ReconnectingMetaStoreClientFactory.java
courtsVII: Things had to change so it can work for both projects, also a lot of these reflection MetaStoreClient classes didn't have tests, good time to add them IMO

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/MetaStoreUriNormaliser.java
courtsVII: I'm closing this PR, it doesn't need attention, I am just keeping this branch up to date so I don't have to do a massive context switch and make design decisions I regret when I have to change the project to use new library

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/metastore/ReconnectingMetaStoreClientFactory.java
courtsVII: Also name clashes, Circus Train and Waggle Dance classes have the same names that weren't descriptive, this is a MetaStoreClient that reconnects if it is disconnected -> ReconnectingMetaStoreClient

PR #81, opened 2018-06-08
using dummy client if connection unavailable

### LINKED ISSUE #80: Waggle Dance fails to serve when destination bastion is down/unreachable
**Issue:**
Waggle Dance federates to primary Metasotre `A` and foreign Metastore `B` (over SSH Tunnel through bastion host `BH`).
If for any reason, `BH` is down/unreachable or disallow SSH connection then this cause whole WD to stops serving request even for Metastore `A`.

**Excepted Behaviour:**
WD to handle the above situation gracefully and continue to serve rest of the Metastores.

**Current Behaviour:**
WD fails to serve all requests.

patduin:
fixes #80

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,11 @@
  */
 package com.hotels.bdp.waggledance.mapping.model;
 
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
+import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -46,7 +51,12 @@ public class MetaStoreMappingFactoryImpl implements MetaStoreMappingFactory {
   }
 
   private CloseableThriftHiveMetastoreIface createClient(AbstractMetaStore metaStore) {
-    return metaStoreClientFactory.newInstance(metaStore);
+    try {
+      return metaStoreClientFactory.newInstance(metaStore);
+    } catch (Exception e) {
+      LOG.error("Can't create a client for metastore '{}':", metaStore.getName(), e);
+      return newUnreachableMetatstoreClient(metaStore);
+    }
   }
 
   @Override
@@ -63,4 +73,34 @@ public class MetaStoreMappingFactoryImpl implements MetaStoreMappingFactory {
     return prefixNamingStrategy.apply(federatedMetaStore);
   }
 
+  private CloseableThriftHiveMetastoreIface newUnreachableMetatstoreClient(AbstractMetaStore metaStore) {
+    return (CloseableThriftHiveMetastoreIface) Proxy.newProxyInstance(getClass().getClassLoader(),
+        new Class[] { CloseableThriftHiveMetastoreIface.class },
+        new UnreachableMetastoreClientInvocationHandler(metaStore.getName()));
+  }
+
+  /**
+   * Handler that refuses to be open and will throw exceptions for any of the methods, serves as a dummy client if the
+   * real one can't be created due to connection (i.e. tunneling) issues.
+   */
+  private static class UnreachableMetastoreClientInvocationHandler implements InvocationHandler {
+
+    private final String name;
+
+    private UnreachableMetastoreClientInvocationHandler(String name) {
+      this.name = name;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+      switch (method.getName()) {
+      case "isOpen":
+        return false;
+      case "close":
+        return null;
+      default:
+        throw new TException("Metastore `" + name + "` unavailable");
+      }
+    }
+  }
 }
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
@@ -99,7 +99,7 @@ public class MetaStoreMappingFactoryImpl implements MetaStoreMappingFactory {
       case "close":
         return null;
       default:
-        throw new TException("Metastore `" + name + "` unavailable");
+        throw new TException("Metastore '" + name + "' unavailable");
       }
     }
   }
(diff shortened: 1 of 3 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -63,4 +73,34 @@ public String prefixNameFor(AbstractMetaStore federatedMetaStore) {
...
+
+    private UnreachableMetastoreClientInvocationHandler(String name) {
+      this.name = name;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+      switch (method.getName()) {
+      case "isOpen":
+        return false;
```
massdosage: Does this mean later on WD will think this thing is closed and try create a new one and attempt a reconnect? I'm just trying to think about the sequence of events here - while the remote is unreachable how often will it check that it's back and then recover?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -63,4 +73,34 @@ public String prefixNameFor(AbstractMetaStore federatedMetaStore) {
...
+      case "close":
+        return null;
+      default:
+        throw new TException("Metastore `" + name + "` unavailable");
```
massdosage: Can we please use ' instead of ` ?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImplTest.java
```diff
@@ -96,4 +97,18 @@ public void connectionLost() throws Exception {
...
+    MetaStoreMappingFactoryImpl factory = new MetaStoreMappingFactoryImpl(prefixNamingStrategy,
+        closeableThriftHiveMetastoreIfaceClientFactory, accessControlHandlerFactory);
+    AbstractMetaStore federatedMetaStore = newFederatedInstance("fed1", thrift.getThriftConnectionUri());
+    when(closeableThriftHiveMetastoreIfaceClientFactory.newInstance(federatedMetaStore))
+        .thenThrow(new RuntimeException("Cannot create client"));
+
+    MetaStoreMapping mapping = factory.newInstance(federatedMetaStore);
+    assertThat(mapping, is(notNullValue()));
+    assertThat(mapping.isAvailable(), is(false));
+  }
```
massdosage: If the idea is that at some point the failure can be recovered from it would be good to have a test showing the sequence of events and the recovery happening.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
patduin: User(hive/spark client) opens a session does a request for the duration of that session this client lives. So if you do a query (show databases) it won't show databases for the broken metastore. If you do a query on the specific database in that metastore (prefix and all) it will fail like hive would fail if you try to query a database that doesn't exist. 
If you close your client and reconnect it will check all metastores again.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
patduin: oh `yes`

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImplTest.java
patduin: you need to create a new instance, that's a bit silly to test in this method

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
massdosage: OK, makes sense, so for a call to a Thrift method it will create one of these, the specific method call will throw the below exception which WD will log and ignore and this will carry on until you reconnect at which point it will attempt to connect again etc.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImplTest.java
massdosage: OK, yeah, I understand the lifecycle better now.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -63,4 +73,34 @@ public String prefixNameFor(AbstractMetaStore federatedMetaStore) {
...
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+      switch (method.getName()) {
+      case "isOpen":
+        return false;
+      case "close":
+        return null;
+      default:
+        throw new TException("Metastore '" + name + "' unavailable");
```
massdosage: Can you write a test for this behaviour?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
patduin: yes, you have to reconnect. If you are doing multiple queries inside the hive-client it will use the disconnected one.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
massdosage: Ah, OK, right, makes sense. Thanks for the explanation.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -63,4 +73,34 @@ public String prefixNameFor(AbstractMetaStore federatedMetaStore) {
...
+    return (CloseableThriftHiveMetastoreIface) Proxy.newProxyInstance(getClass().getClassLoader(),
+        new Class[] { CloseableThriftHiveMetastoreIface.class },
+        new UnreachableMetastoreClientInvocationHandler(metaStore.getName()));
+  }
+
+  /**
+   * Handler that refuses to be open and will throw exceptions for any of the methods, serves as a dummy client if the
+   * real one can't be created due to connection (i.e. tunneling) issues.
+   */
+  private static class UnreachableMetastoreClientInvocationHandler implements InvocationHandler {
```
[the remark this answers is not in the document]
patduin: yes it has no business being used anywhere else

PR #85, opened 2018-06-14
handling parse errors

### LINKED ISSUE #83: handle view parse error silently
We are hitting an Hive bug in parsing the views https://issues.apache.org/jira/browse/HIVE-19896
This affects doing among others: `desc table`, `show create table` commands via Waggle Dance (only affects federated metastores, primary metastore is ok).

We should ignore the errors so the client is not affected

patduin:
fixes #83 
The view text we get back is not parseable due to https://issues.apache.org/jira/browse/HIVE-19896
Catching the exceptions so we won't fail the client connects for (as far as I can see) all practical purposes (show create table. desc table) we get the viewExpandedText back in hive-cli. Other queries that imply fetching the table object probably don't use this field (very hard to be sure).


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -77,14 +83,28 @@ public DatabaseMappingImpl(MetaStoreMapping metaStoreMapping, QueryMapping query
...
+            queryMapping.transformOutboundDatabaseName(metaStoreMapping, table.getViewExpandedText()));
+      } catch (WaggleDanceException e) {
+        log.debug("Error while transforming databaseName in ViewExpandedText, keeping original", e);
+      }
     }
     if (table.isSetViewOriginalText()) {
-      table.setViewOriginalText(
-          queryMapping.transformOutboundDatabaseName(metaStoreMapping, table.getViewOriginalText()));
+      try {
+        log.debug("Transforming ViewOriginalTex: {}", table.getViewOriginalText());
```
massdosage: ViewOriginalTex -> ViewOriginalText

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -77,14 +83,28 @@ public DatabaseMappingImpl(MetaStoreMapping metaStoreMapping, QueryMapping query
...
+        table.setViewOriginalText(
+            queryMapping.transformOutboundDatabaseName(metaStoreMapping, table.getViewOriginalText()));
+      } catch (WaggleDanceException e) {
+        log.debug("Error while transforming databaseName in ViewOriginalTex, using ViewExpandedText if available", e);
```
massdosage: ViewOriginalTex -> ViewOriginalText

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -77,14 +83,28 @@ public DatabaseMappingImpl(MetaStoreMapping metaStoreMapping, QueryMapping query
...
+        if (table.isSetViewExpandedText()) {
+          table.setViewOriginalText(table.getViewExpandedText());
```
massdosage: Why do we do this instead of just leaving the original text as it was?

patduin: I've discussed this with @teabot, since the original text is not parseable and the db's in it are wrong we felt it might be better to add the parseable and correctly transformed expanded view. Hard to say what is better, depends on how this value is used. I'm happy to keep original as well.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
patduin: fixed them

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -77,14 +83,28 @@ public DatabaseMappingImpl(MetaStoreMapping metaStoreMapping, QueryMapping query
...
-      table.setViewOriginalText(
-          queryMapping.transformOutboundDatabaseName(metaStoreMapping, table.getViewOriginalText()));
+      try {
+        log.debug("Transforming ViewOriginalText: {}", table.getViewOriginalText());
+        table.setViewOriginalText(
+            queryMapping.transformOutboundDatabaseName(metaStoreMapping, table.getViewOriginalText()));
+      } catch (WaggleDanceException e) {
+        log.debug("Error while transforming databaseName in ViewOriginalText, using ViewExpandedText if available", e);
+        if (table.isSetViewExpandedText()) {
+          table.setViewOriginalText(table.getViewExpandedText());
```
ddcprg: not sure about this, is this a temporary fix? can you put a comment explaining why this is done here?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/QueryMapping.java
```diff
@@ -15,13 +15,16 @@
...
 
   /**
    * @param metaStoreMapping
    * @param query
    * @return query where database names are correctly mapped using metaStoreMapping. example 'select id from db.table'
    *         -> 'select id from waggle_db.table'
+   * @throws WaggleDanceException when the transform could not be done.
    */
-  String transformOutboundDatabaseName(MetaStoreMapping metaStoreMapping, String query);
+  String transformOutboundDatabaseName(MetaStoreMapping metaStoreMapping, String query) throws WaggleDanceException;
```
ddcprg: is `WaggleDanceException` not a `RuntimeException`?

patduin: it is but I wanted to make it part of the Interface contract so users of it know what can be expected to catch

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
patduin: The fix is permanent because we can always hit the case where the QueryMapping fails and we shouldn't fail the client. But I can add a comment with the link to the Hive bug (https://issues.apache.org/jira/browse/HIVE-19896)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/QueryMapping.java
ddcprg: is documented, it doesn't need to be in the method signature

PR #86, opened 2018-06-14
Using text replacements instead of tokensToString

### LINKED ISSUE #84: Waggle Dance adds double backticks in certains places when transforming view queries.
example: 
```
select a from (select b from c) as `d.c`
```
Is transformed by our transformation into:
```
select a from (select b from c) as ``d.c``
```
`show create table` on the view with that select will return the double backticked value.

patduin:
fixes #84


waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMappingTest.java
```diff
@@ -37,13 +38,34 @@ public void transformOutboundDatabaseName() {
...
-        queryMapping.transformOutboundDatabaseName(metaStoreMapping, query));
+            + "db2.table2 alias2\n"
+            + "ON alias1.field1 = alias2.field2"));
+  }
+
+  @Test
+  public void transformOutboundDatabaseNameAliasWithBackTicks() {
+    ASTQueryMapping queryMapping = ASTQueryMapping.INSTANCE;
+    MetaStoreMapping metaStoreMapping = new MetaStoreMappingImpl(PREFIX, "mapping", null, null);
+
```
courtsVII: Replace the HCom specific names with something more generic

patduin: ah good point!

teabot: +1

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
```diff
@@ -46,35 +55,65 @@ public String transformOutboundDatabaseName(MetaStoreMapping metaStoreMapping, S
...
+          dbNameTokens.add(dbNameNode);
         }
         // Otherwise TOK_TABNAME node only has one child which contains just the table name
       }
     }
+    return dbNameTokens;
+  }
+
+  private boolean childrenAreIdentifiers(ASTNode current) {
+    for (Node child : current.getChildren()) {
```
teabot: Use `getChildren()`? You have the `Node` → `ASTNode` casting in to places currently: L101,113. I suspect you should use one approach or the other.

patduin: yup good idea

Issue #93, opened 2018-06-19
Enable custom ThriftHiveMetastore.IFace implementations


courtsVII:
Enable configurable pluggable implementations of the ThriftHiveMetastore.IFace interface on a per (primary or federated) metastore basis. 

Example of how the configuration could look like
```
primary-meta-store:
  access-control-type: READ_ONLY
  database-prefix: ''
  name: primary
  closeable-iface: com.company.foo.bar.MyImpl
  #com.company.foo.bar.MyImpl is a custom 
  #implementation of the CloseableThriftHiveMetastoreIface  that can forward
  #requests to some service in the background and adapt them to the Hive interface 
  #e.g. federate Glue onto hive, federate BigQuery onto hive etc by using this
  #interface as an adapter from your data store to the 
  #CloseableThriftHiveMetastoreIface  interface 

federated-meta-stores:
- access-control-type: READ_ONLY
  database-prefix: waggle_test_
  mapped-databases: []
  name: secondary
  remote-meta-store-uris: thrift://waggle-dance:48869
  #No closeable-iface implementation is specified so it defaults to Thrift 

  #Thrift Hive will be federated onto your custom implementation

```

massdosage:
Closing for now, we'll re-open this if/when we have a requirement that touches this code path.

PR #94, opened 2018-06-22
Waggle Dance Plugins: Allow jars to be added to the classpath without mvn dependency

courtsVII:
…aven dependency

### DIFF
--- a/waggle-dance-core/pom.xml
+++ b/waggle-dance-core/pom.xml
@@ -287,6 +287,7 @@
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot-maven-plugin.version}</version>
         <configuration>
+          <layout>ZIP</layout>
           <requiresUnpack>
             <!-- We are hitting an issue introduced in hive-common-2.1.1 and fixed in https://issues.apache.org/jira/browse/HIVE-17155
               The HiveConf class tries loading conf (e.g. hive-site.xml) from the uberjar and fails. Unpacking it with this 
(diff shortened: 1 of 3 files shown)


courtsVII:
resolves #93

waggle-dance/src/main/service/waggle-dance.conf
```diff
@@ -1,4 +1,4 @@
 #Spring boot service conf: see http://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html#deployment-script-customization-conf-file
 LOG_FOLDER=/var/log/waggle-dance/
-JAVA_OPTS="-Xmx12g -XX:+UseG1GC -XX:G1ReservePercent=15 -Dlog4j.configurationFile=/opt/waggle-dance/conf/log4j2.xml -Dlogging.config=/opt/waggle-dance/conf/log4j2.xml"
+JAVA_OPTS="-Xmx12g -XX:+UseG1GC -XX:G1ReservePercent=15 -Dloader.path=lib,/opt/waggle-dance/lib/ -Dlog4j.configurationFile=/opt/waggle-dance/conf/log4j2.xml -Dlogging.config=/opt/waggle-dance/conf/log4j2.xml"
```
ddcprg: does this not work with the jar layout?

courtsVII: Not with how we have it prior to this PR

waggle-dance-core/pom.xml
```diff
@@ -287,6 +287,7 @@
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot-maven-plugin.version}</version>
         <configuration>
+          <layout>ZIP</layout>
```
massdosage: Can you provide an explanation of what exactly this achieves? Perhaps you could unzip the .tgz that we produce with and without this change and run `tree` over the files/folders so we can see? (I'd try this out myself but I don't have the time right now!)

massdosage:
Not clear why this is needed as no other users have reported this as an issue.

PR #95, opened 2018-06-25
allow users to write to federated metastores

### LINKED ISSUE #87: enable READ_WRITE support for federated metastores
As a user of WD
I like to run DDL on federated metastores
So that I can run ETL againt primary metastore and update federated central datalake.

Acceptance Criteria:
  ability to configure access-control-type option for federated metastores with atleast following possible values READ_ONLY, READ_WRITE.

courtsVII:
resolves #87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [TBD]
+### Changed
+* Enable federated metastore access-control-type to be configured to READ_ONLY, READ_AND_WRITE_AND_CREATE, READ_AND_WRITE_ON_DATABASE_WHITELIST or READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST
+
+
 ## [2.3.7] 2018-06-19
 ### Fixed
 * Silently handling parse errors in views. See [#83](https://github.com/HotelsDotCom/waggle-dance/issues/83).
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## [TBD]
+### Changed
+* Enable federated metastore access-control-type to be configured to READ_ONLY, READ_AND_WRITE_AND_CREATE, READ_AND_WRITE_ON_DATABASE_WHITELIST or READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST
```
patduin: we probably need to decide how to handle `create database`, couple of options:
1) only allow READ_ONLY and READ_AND_WRITE_ON_DATABASE_WHITELIST (perhaps we add a READ_AND_WRITE?)
2) we agree on a strategy for creating db's. PrefixMode we follow prefixes (so we create in remote if it has the correct prefix (transforming the db name along the way), ManualMode we only create db's in primary.

NOTE: and access control strategy that says WRITE allows for `drop database` statements to be performed.

README.md
```diff
@@ -176,6 +176,24 @@ This is controlled by the property: `primary-meta-store.access-control-type` It
 | `READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST`     | Reads are allowed, writes are allowed on database names listed in the `primary-meta-store.writable-database-white-list` property, creating new databases is allowed and they are added to the white-list automatically. |
 | `READ_AND_WRITE_ON_DATABASE_WHITELIST`                | Reads are allowed, writes are allowed on database names listed in the `primary-meta-store.writable-database-white-list` property, creating new databases is not allowed. |
```
patduin: this needs to be updated as the writable-database-white-list can now also be set on the federated metastore correct?

CHANGELOG.md
courtsVII: Should we do this in another ticket or tackle it now? I feel like we will want some consensus on how we should handle it

PR #104, opened 2018-07-20
Issue 100

### LINKED ISSUE #100: Increase code coverage (2)
_As a developer_
_I'd like to be confident my tests can detect refactoring issues_
_So I can easily make code changes_

**Acceptance criteria**
Code coverage should be at least 80%

Below is a list of classes which could benefit from some extra testing and should be the focus of this ticket. They are arranged (more or less) in increasing test complexity:

ValidationError
FederatedMetaStore
PrimaryMetaStore
MetastoreUnavailableException
TunnelableFactorySupplier
TunnelingMetaStoreClientFactory
NotifyingFederationService
AdvancedPropertyUtils

PrefixBasedDatabaseMappingService
WaggleDanceConfiguration
YamlStorageConfiguration (trivial)
MonitoringConfiguration

FederatedHMSHandlerFactory
FederatedHMSHandler

AnanaMJ:
Fixes #100


waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.validation.ObjectError;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ValidationErrorTest {
+
+  private @Mock Errors errors;
+  private @Mock ValidationError validationError;
+
+  @Test
+  public void testBuilder() {
```
massdosage: Prefixing test methods with `test` was required in earlier versions of JUnit but is no longer needed since we have the annotation. Generally we try avoid this naming convention. I see a few other tests in this package do that but most of them don't. So please rename the methods to be a bit more descriptive as to what they're trying to test. This one could be something like `builderTypical()`?

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    ValidationError result = ValidationError.builder().build();
+    assertThat(result.getErrorMessage(), is("Validation failed"));
```
massdosage: This is a bit brittle depending on an internally declared String but I can't think of a better suggestion other than over-complicating things and introducing a package viewable constant etc. which isn't worth the effort. So it's fine to leave it like this.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    assertThat(result.getErrors().size(), is(0));
+  }
+
+  @Test
+  public void testBuilderErrors() {
+    ValidationError result = ValidationError.builder(errors).build();
+    assertThat(result.getErrorMessage(), is("Validation failed: 0 error(s)"));
```
massdosage: Seeing this actually makes one question why the class lets one pass in empty errors. Anyway, it's fine to leave for now as this is how it works but it suggests we could have built it better so that if no errors are passed in you don't get a "failed" message.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    assertThat(result.getErrors().size(), is(0));
+  }
+
+  @Test
+  public void testMultipleBuilderErrors() {
+    List<ObjectError> objectError = new ArrayList<>();
```
massdosage: objectError -> objectErrors

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    objectError.add(new ObjectError("errorOne", "Description one"));
+    objectError.add(new ObjectError("errorTwo", "Description two"));
+
+    when(errors.getErrorCount()).thenReturn(objectError.size());
+    when(errors.getAllErrors()).thenReturn(objectError);
+
+    List<String> expected = new ArrayList<>();
```
massdosage: We generally aim to have declaration as close as possible to use, so you should move this down to below line 65.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    expected.add("Description one");
```
massdosage: The downside of doing this is that it you change the description above when you create the ObjectError, you also need to change it here. An alternative would be to have local variables objectError1, objectError2 which are added to the list and here you could do `expected.add(objectError1.getDescription())`

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    assertThat(result.getErrorMessage(), is("Validation failed: 2 error(s)"));
+  }
+
+  @Test
+  public void testAddValidationError() {
+    ValidationError validationError = ValidationError.builder().build();
+    validationError.addValidationError("Another error");
+
+    List<String> expectedErrors = new ArrayList<String>();
+    expectedErrors.add("Another error");
```
massdosage: Similar, you could declare this as a  String and then reference it twice rather than have 2 Strings with the same value.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/FederatedMetaStoreConstructorTest.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.api.model;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+public class FederatedMetaStoreConstructorTest extends AbstractMetaStoreTest<FederatedMetaStore> {
```
massdosage: Not sure if you wanted me to review this yet but why aren't these tests just in the FederatedMetaStore test?

waggle-dance-core/pom.xml
```diff
@@ -11,8 +11,8 @@
 
   <properties>
     <spring-boot-maven-plugin.version>1.5.7.RELEASE</spring-boot-maven-plugin.version>
-    <mockito.version>1.10.19</mockito.version>
-    <powermock.version>1.6.4</powermock.version>
+    <mockito.version>2.8.47</mockito.version>
```
massdosage: Hopefully we'll get @andreeapad's PR with the version changes merged into master soon and we can then merge that in here and get all the version numbers consistent. Until then this is fine.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.springframework.validation.Errors;
+import org.springframework.validation.ObjectError;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ValidationErrorTest {
```
massdosage: This looks good now, nice work.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/FederatedMetaStoreTest.java
```diff
@@ -77,4 +79,22 @@ public void toJson() throws Exception {
...
+    mappedDatabases.add("database");
+    metaStore.setMappedDatabases(mappedDatabases);
+    assertThat(metaStore.getMappedDatabases(), is(mappedDatabases));
+  }
+
+  @Test
+  public void nonEmptyConstructor() {
+    String storeName = "name";
+    String storeUri = "remoteMetaStoreUris";
+    FederatedMetaStore store = new FederatedMetaStore(storeName, storeUri);
```
massdosage: OK, now I see what you were trying to do earlier as these tests extend a base class which be default calls the no-args constructor. I think what you've done now is better than introducing a new test class.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/FederatedMetaStoreTest.java
```diff
@@ -1,5 +1,5 @@
 /**
```
massdosage: This test looks good, we can consider it done.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStoreTest.java
```diff
@@ -19,6 +19,8 @@
 import static org.hamcrest.Matchers.notNullValue;
```
massdosage: This test looks good, we can consider it done.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/MetastoreUnavailableExceptionTest.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+public class MetastoreUnavailableExceptionTest {
+
+  @Test
+  public void messageError() {
+    MetastoreUnavailableException exception = new MetastoreUnavailableException("error");
```
massdosage: Same as my comment on the other test about putting the value into a string that you can assert later instead of declaring it twice.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelableFactorySupplierTest.java
```diff
@@ -110,4 +110,10 @@ public void incorrectStrictHostKeyCheckingSetting() {
     supplier.get(hiveConf);
   }
 
+  @Test(expected = IllegalArgumentException.class)
```
massdosage: I don't understand why this would throw the exception? "no" is valid input is it not? Or have you not got up to here yet?

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/FederatedMetaStoreTest.java
AnanaMJ: Yeah, on Friday I didn't know how to test the other constructor, but I got some help from @teabot to understand how to do it :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelableFactorySupplierTest.java
AnanaMJ: That's true, it doesn't throw an exception! I changed it :)

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStoreTest.java
```diff
@@ -30,6 +32,11 @@
 
 public class PrimaryMetaStoreTest extends AbstractMetaStoreTest<PrimaryMetaStore> {
 
+  String name = "name";
```
massdosage: Make all of these private. We generally aim to keep the access level of everything as closed as possible.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/MetastoreUnavailableExceptionTest.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.hotels.bdp.waggledance.client;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+public class MetastoreUnavailableExceptionTest {
```
massdosage: This looks good.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -76,4 +77,10 @@ public void newInstanceWithoutTunneling() throws Exception {
     verifyZeroInteractions(tunnelableFactorySupplier);
   }
 
+  @Test
+  public void newInstanceWithoutTunnelingNorSupplier() {
+    TunnelingMetaStoreClientFactory clientFactory = new TunnelingMetaStoreClientFactory();
+    clientFactory.newInstance(hiveConf, "test", 10);
```
massdosage: What are you testing here? Just that no exceptions are being thrown? (there are no assertions)

AnanaMJ: This is supposed to test the first constructor of the class that doesn't need any parameters.

massdosage: Yes, but you're not doing anything after that, is there anything you can assert? Maybe based on the values you've passed in to check they're being used as expected?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
```diff
@@ -112,12 +80,13 @@ public CloseableThriftHiveMetastoreIface newInstance(HiveConf hiveConf, String n
     TunnelableFactory<CloseableThriftHiveMetastoreIface> tunnelableFactory = tunnelableFactorySupplier
         .get(localHiveConf);
 
-    LOG.info("Metastore URI {} is being proxied through {}", hiveConf.getVar(HiveConf.ConfVars.METASTOREURIS),
-        localHiveConf.getVar(HiveConf.ConfVars.METASTOREURIS));
+    LOG
+        .info("Metastore URI {} is being proxied through {}", hiveConf.getVar(HiveConf.ConfVars.METASTOREURIS),
```
courtsVII: Formatting

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplier.java
```diff
@@ -0,0 +1,28 @@
+package com.hotels.bdp.waggledance.client.tunneling;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+
+import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIface;
+import com.hotels.bdp.waggledance.client.DefaultMetaStoreClientFactory;
+import com.hotels.bdp.waggledance.client.MetaStoreClientFactory;
+import com.hotels.hcommon.ssh.TunnelableSupplier;
+
```
courtsVII: Make Package Private

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplier.java
```diff
@@ -0,0 +1,28 @@
...
+public class HiveMetaStoreClientSupplier implements TunnelableSupplier<CloseableThriftHiveMetastoreIface> {
+  private final MetaStoreClientFactory factory;
+  private final HiveConf hiveConf;
+  private final String name;
+  private final int reconnectionRetries;
+
+  public HiveMetaStoreClientSupplier(HiveConf hiveConf, String name, int reconnectionRetries) {
```
courtsVII: Ditto

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/MetastoreClientMethodChecker.java
```diff
@@ -0,0 +1,17 @@
+package com.hotels.bdp.waggledance.client.tunneling;
+
+import java.lang.reflect.Method;
+
+import com.hotels.hcommon.ssh.MethodChecker;
+
```
courtsVII: Package Private

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactoryTest.java
AnanaMJ: So far I have tried to get the configuration values from the object returned by `newInstance`, but it does not let me access config values: `For security reasons, the config key com.hotels.bdp.waggledance.client.ssh.localhost cannot be accessed`.
This happens for all the variables in the config.

Also, the client name and number of retries are passed to a private class within `DefaultMetaStoreClientFactory` so I don't think those are accessible either.

One thing that I can check for is that the object returned from `clientFactory.newInstace(...)` is not null. Is there anything else that I'm missing?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
AnanaMJ: I can't change this :(
Eclipse formatter just splits it like that and if I remove it manually, the next time I save the file, Eclipse puts the line break back there.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,154 @@
...
+    tunnelingMetaStoreClientFactory.newInstance(hiveConf, name, reconnectionRetries);
+    ArgumentCaptor<String> stringArgument = ArgumentCaptor.forClass(String.class);
+    ArgumentCaptor<Integer> intArgument = ArgumentCaptor.forClass(Integer.class);
+    verify(localHiveConfFactory).createLocalHiveConf(stringArgument.capture(), intArgument.capture(), eq(hiveConf));
+    assertThat(stringArgument.getValue(), is(localHost));
+    assertThat(intArgument.getValue(), is(localPort));
+  }
+
+  @Test
+  public void tunnelableFactorySupplierUsesCorrectParameters() {
```
patduin: These tests are too fragmented, you making a test to verify that a line is called that doesn't add much imo. Look at the number of conditionals in the code thats the number of tests you should have to get full coverage.
Got one if, means you need two tests as there are two possible paths code can take and you want to run both in your test.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,154 @@
...
+    String remoteHost = uri.getHost();
+    int remotePort = uri.getPort();
+    tunnelingMetaStoreClientFactory.newInstance(hiveConf, name, reconnectionRetries);
+    verify(tunnelableFactory)
+        .wrap(eq(hiveMetaStoreClientSupplier), eq(tunnelingMetaStoreClientFactory.METHOD_CHECKER), eq(localHost),
+            any(Integer.class), eq(remoteHost), eq(remotePort));
+  }
+
+  @Test
+  public void newInstanceWithoutTunneling() throws Exception {
```
patduin: use *asserts* for validating your test if possible. `verify` is to test delegation you should only need to use it if you have some kind of method that doesn't return anything and you want to make sure something else is called correctly. 
In this case the class under test returns something, so all test should end with an assertion of the object.

example instead of:
```
class_under_test.create(someFactory);
verify(factory).methodCalled("string");
```
You mock the behavior of `factory` like so:
```
when(factory.methodCall("string").thenReturn("anotherString");
String result = class_under_test.create(someFactory);
assertThat(result, is("anotherString");
```

You are not validating the returned object at all which is bad as that's the thing that counts the most. Mock the expected behavior from the classes you inject, that documents what you expect from them better then `verify` they are called, its more important mock what they return.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
AnanaMJ: Some of these tests are to prevent previous bugs from happening again. @courtsVII told me that at one point in the past, the code was changed and bugs were introduced because `hiveConf` was used instead of `localConf`. So while less tests would be enough for this class, I thought that being more defensive was the right approach.

patduin: sure there are always exception but I think you went a bit too far, have a look at my other comment and if you rewrite some of the tests following those suggestions see if these small ones are actually adding anything or needed.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    assertThat(result.getErrorMessage(), is("Validation failed: 0 error(s)"));
+    assertThat(result.getErrors().size(), is(0));
+  }
+
+  @Test
+  public void multipleBuilderErrors() {
+    ObjectError objectError1 = new ObjectError("errorOne", "Description one");
+    ObjectError objectError2 = new ObjectError("errorTwo", "Description two");
+
+    List<ObjectError> objectErrors = new ArrayList<>();
```
teabot: for brevity:
```
List<ObjectError> objectErrors = Arrays.asList(objectError1, objectError2);
```

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    objectErrors.add(objectError1);
+    objectErrors.add(objectError2);
+
+    when(errors.getErrorCount()).thenReturn(objectErrors.size());
+    when(errors.getAllErrors()).thenReturn(objectErrors);
+
+    ValidationError result = ValidationError.builder(errors).build();
+
+    List<String> expected = new ArrayList<>();
```
teabot: as above

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/ValidationErrorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    assertThat(result.getErrorMessage(), is("Validation failed: 2 error(s)"));
+  }
+
+  @Test
+  public void addValidationError() {
+    String errorMessage = "Another error";
+    ValidationError validationError = ValidationError.builder().build();
+    validationError.addValidationError(errorMessage);
+
+    List<String> expectedErrors = new ArrayList<String>();
```
teabot:
```
List<String> expectedErrors = Collections.singletonList(errorMessage);
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -89,15 +89,16 @@ private void init(List<AbstractMetaStore> federatedMetaStores) {
   private void add(AbstractMetaStore federatedMetaStore) {
     MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(federatedMetaStore);
     if (federatedMetaStore.getFederationType() == PRIMARY) {
+      System.out.println("adding to mapping: " + federatedMetaStore.getName());
```
teabot: As a rule, we very, very rarely want to use `System.out`. Direct this to the logging framework instead.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplierTest.java
```diff
@@ -0,0 +1,50 @@
...
+import com.hotels.bdp.waggledance.client.MetaStoreClientFactory;
+import com.hotels.beeju.ThriftHiveMetaStoreJUnitRule;
+
+@RunWith(MockitoJUnitRunner.class)
+public class HiveMetaStoreClientSupplierTest {
+
+  public @Rule ThriftHiveMetaStoreJUnitRule metastore = new ThriftHiveMetaStoreJUnitRule();
+
+  private @Mock MetaStoreClientFactory factory;
+  private HiveConf hiveConf;
```
teabot: Does this need to be a member variable at this point? Always try to decrease scope.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/LocalHiveConfFactoryTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import com.hotels.beeju.ThriftHiveMetaStoreJUnitRule;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LocalHiveConfFactoryTest {
+  public @Rule ThriftHiveMetaStoreJUnitRule metastore = new ThriftHiveMetaStoreJUnitRule();
+
+  private HiveConf hiveConf;
```
teabot: scope

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,154 @@
...
+    tunnelingMetaStoreClientFactory.newInstance(hiveConf, name, reconnectionRetries);
+    verify(tunnelableFactory)
+        .wrap(captor.capture(), eq(tunnelingMetaStoreClientFactory.METHOD_CHECKER), eq("my-machine"), anyInt(),
+            eq("localhost"), eq(metastore.getThriftPort()));
+  }
+
+  @Test
+  public void localHiveConfigUsesCorrectParameters() {
+    tunnelingMetaStoreClientFactory.newInstance(hiveConf, name, reconnectionRetries);
+    ArgumentCaptor<String> stringArgument = ArgumentCaptor.forClass(String.class);
```
teabot: You can also create `ArgumentCaptor` with the `@Captor` annotation, might not be appropriate here but for the record:
```
private @Captor ArgumentCaptor<String> stringArgument;
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: This class is a work in progress right now, so all the system.out commands will be removed after the testing is done for it :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplierTest.java
AnanaMJ: This seems to be needed, otherwise an assertion error is thrown, but the other one can be removed.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/LocalHiveConfFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.client.tunneling;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+
+public class LocalHiveConfFactory {
+
+  HiveConf createLocalHiveConf(String localHost, int localPort, HiveConf hiveConf) {
```
massdosage: If it's a factory then this should be called `newInstance`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/LocalHiveConfFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.client.tunneling;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+
+public class LocalHiveConfFactory {
```
massdosage: Does this class need to be `public`? The method inside isn't...

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/LocalPortFactory.java
```diff
@@ -0,0 +1,18 @@
+package com.hotels.bdp.waggledance.client.tunneling;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+import com.hotels.hcommon.ssh.SshException;
+
+class LocalPortFactory {
```
massdosage: This looks a bit odd as a factory, couldn't it just be a method in the class where you use it from?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -96,8 +96,8 @@ private void add(AbstractMetaStore federatedMetaStore) {
       Whitelist mappedDbWhitelist = null;
       if (FederatedMetaStore.class.isAssignableFrom(federatedMetaStore.getClass())) {
         mappedDbWhitelist = getWhitelistedDatabases((FederatedMetaStore) federatedMetaStore);
+        mappedDbByPrefix.put(metaStoreMapping.getDatabasePrefix(), mappedDbWhitelist);
```
massdosage: You're 100% sure that we never need an entry in the map for a prefix against a null whitelist? By moving that here this behaviour has now changed which might be fine, I'm just checking there's definitely a reason for this.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplierTest.java
massdosage: I don't see any reason why the `HiveConf` declaration can't be moved down to line 42.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplierTest.java
```diff
@@ -0,0 +1,50 @@
...
+  @Test
+  public void getMetaStoreClientFactoryInstance() {
+    String name = "test";
+    int reconnectionRetries = 10;
+    hiveConf = metastore.conf();
+    HiveMetaStoreClientSupplier supplier = new HiveMetaStoreClientSupplier(factory, hiveConf, name,
+        reconnectionRetries);
+    supplier.get();
+    verify(factory).newInstance(hiveConf, name, reconnectionRetries);
+
```
massdosage: remove blank line

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/LocalHiveConfFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import com.hotels.beeju.ThriftHiveMetaStoreJUnitRule;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LocalHiveConfFactoryTest {
+  public @Rule ThriftHiveMetaStoreJUnitRule metastore = new ThriftHiveMetaStoreJUnitRule();
```
massdosage: Blank line before this.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,154 @@
...
+    when(localPortFactory.getLocalPort()).thenReturn(localPort);
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_PRIVATE_KEYS.varname, "private_key");
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_KNOWN_HOSTS.varname, "");
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_STRICT_HOST_KEY_CHECKING.varname, "yes");
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_ROUTE.varname, "user@hop1 -> hop2");
+    hiveConf.set(WaggleDanceHiveConfVars.SSH_LOCALHOST.varname, localHost);
+
+    tunnelingMetaStoreClientFactory = new TunnelingMetaStoreClientFactory(tunnelableFactorySupplier,
+        metaStoreClientFactory, localHiveConfFactory, hiveMetaStoreClientSupplierFactory, localPortFactory);
+
```
massdosage: remove blank line

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfigurationTest.java
```diff
@@ -30,6 +30,8 @@
```
massdosage: This test looks fine.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,154 @@
...
+import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIface;
+import com.hotels.bdp.waggledance.client.MetaStoreClientFactory;
+import com.hotels.bdp.waggledance.client.WaggleDanceHiveConfVars;
+import com.hotels.beeju.ThriftHiveMetaStoreJUnitRule;
+import com.hotels.hcommon.ssh.TunnelableFactory;
+import com.hotels.hcommon.ssh.TunnelableSupplier;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TunnelingMetaStoreClientFactoryTest {
```
massdosage: This looks like it still has some outstanding comments from @patduin that need to be actioned or replied to.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/conf/YamlStorageConfigurationTest.java
```diff
@@ -0,0 +1,23 @@
+package com.hotels.bdp.waggledance.conf;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+public class YamlStorageConfigurationTest {
```
massdosage: This test looks fine.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
```diff
@@ -355,4 +360,127 @@ public void panopticOperationsHandlerSetUgi() throws Exception {
...
+    MetaStoreMapping unavailableMapping = mockNewMapping(false, "name2_");
+    when(metaStoreMappingFactory.newInstance(unavailableMetastore)).thenReturn(unavailableMapping);
+
+    service = new PrefixBasedDatabaseMappingService(metaStoreMappingFactory,
+        Arrays.asList(federatedMetastore, unavailableMetastore), queryMapping);
+    String databaseName = "some_unknown_prefix_db";
+    try {
+      service.databaseMapping("some_unknown_prefix_db");
+    } catch (NoPrimaryMetastoreException noPrimaryException) {
+      String expectedErrorMessage = "Waggle Dance error no database mapping available tried to map database '"
```
massdosage: It can be quite brittle to assert exception messages as changing the message will break this test and we probably don't care _that_ much about the actual message in this case. I'll leave it up to you whether you want to keep this here or just remove the assertion and the whole try/catch/

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/yaml/AdvancedPropertyUtilsTest.java
```diff
@@ -54,48 +55,87 @@ public void setTransientProperty(char transientProperty) {
...
+    public String getLongPropertyName() {
+      return longPropertyName;
+    }
+  }
+
+  private static class UnreadableTestBean {
+    private String longPropertyName;
+    private char transientProperty;
+
+    @Transient
```
massdosage: Why the need for this annotation? Don't you just need a bean which has a setter with no getter? i.e. the "inverse" of the bean above.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: If we do, then it may be a problem because if `mappedDbWhitelist` is null, then the `put` method throws a NullPointerException. After testing it, I found out that it uses a ConcurrentHashMap which returns that exception if the key or value is null.

massdosage: Oh OK, that makes sense to have it like this then and if all the other tests pass then we should be fine.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunneling/LocalPortFactory.java
AnanaMJ: I had it to mock the local port for TunnelingMetaStoreClientFactoryTest in `localHiveConfigUsesCorrectParameters`. But now I don't think there's much point in doing that because there is already an assertion for other parameters. So I've turned the LocalPortFactory in a private method.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
AnanaMJ: Yes, this class needs a bit more work and I've raised a separate issue to do it after the tests are completed.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/yaml/AdvancedPropertyUtilsTest.java
AnanaMJ: You're right. I don't know why I left the `@Transient` there, but I've removed it now.

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/FederatedMetaStoreTest.java
```diff
@@ -18,6 +18,8 @@
 import static org.hamcrest.CoreMatchers.is;
```
massdosage: This test looks good.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/TunnelingMetaStoreClientFactory.java
massdosage: OK, a bit weird, but let's not worry about this now, we're probably going to need to review some of our formatting rules soon and can check what's going with things like this then.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplierTest.java
```diff
@@ -0,0 +1,48 @@
...
+  public @Rule ThriftHiveMetaStoreJUnitRule metastore = new ThriftHiveMetaStoreJUnitRule();
+
+  private @Mock MetaStoreClientFactory factory;
+
+  @Test
+  public void getMetaStoreClientFactoryInstance() {
+    String name = "test";
+    int reconnectionRetries = 10;
+    HiveConf hiveConf = metastore.conf();
```
massdosage: I'm guessing the code below didn't easily allow you to pass in a mocked HiveConf  so you had to go via a rule to get one?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/LocalHiveConfFactoryTest.java
```diff
@@ -0,0 +1,46 @@
+/**
```
massdosage: This test looks good.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelableFactorySupplierTest.java
```diff
@@ -13,12 +13,15 @@
  * See the License for the specific language governing permissions and
```
massdosage: This test looks good.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/TunnelingMetaStoreClientFactoryTest.java
massdosage: OK, so we'll leave this as is and changes will be done via that issue.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/NotifyingFederationServiceTest.java
```diff
@@ -18,6 +18,7 @@
 import static org.hamcrest.CoreMatchers.is;
```
massdosage: This test looks good.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
```diff
@@ -355,4 +360,118 @@ public void panopticOperationsHandlerSetUgi() throws Exception {
...
+  @Test(expected = NoPrimaryMetastoreException.class)
+  public void noPrimaryMappingThrowsException() {
+    when(metaStoreMappingFactory.newInstance(federatedMetastore)).thenReturn(metaStoreMappingFederated);
+    AbstractMetaStore unavailableMetastore = newFederatedInstance("name2", "thrift:host:port");
+    MetaStoreMapping unavailableMapping = mockNewMapping(false, "name2_");
+    when(metaStoreMappingFactory.newInstance(unavailableMetastore)).thenReturn(unavailableMapping);
+
+    service = new PrefixBasedDatabaseMappingService(metaStoreMappingFactory,
+        Arrays.asList(federatedMetastore, unavailableMetastore), queryMapping);
+    service.primaryDatabaseMapping();
```
massdosage: I can't quite remember how this all works, what is it in the above code that causes the service to not find a primary metastore?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
```diff
@@ -355,4 +360,118 @@ public void panopticOperationsHandlerSetUgi() throws Exception {
...
+  }
+
+  @Test(expected = NoPrimaryMetastoreException.class)
+  public void noPrimaryThrowsExceptionForUnmappedDatabase() {
+    when(metaStoreMappingFactory.newInstance(federatedMetastore)).thenReturn(metaStoreMappingFederated);
+    AbstractMetaStore unavailableMetastore = newFederatedInstance("name2", "thrift:host:port");
```
massdosage: Is `unavailable` the best name for this? The name to me suggests a metastore that is down, perhaps that is the intention here and I'm just missing something.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTest.java
```diff
@@ -0,0 +1,56 @@
...
+public class MonitoringConfigurationTest {
+
+  private final MetricRegistry metricRegistry = new MetricRegistry();
+  private @Mock GraphiteConfiguration graphiteConfiguration;
+
+  private MonitoringConfiguration monitoringConfiguration;
+
+  @Before
+  public void setUp() {
+    monitoringConfiguration = new MonitoringConfiguration();
```
massdosage: Inline on line 27

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTest.java
```diff
@@ -0,0 +1,56 @@
...
+    monitoringConfiguration.setMetricRegistry(metricRegistry);
+    monitoringConfiguration.setGraphiteConfiguration(graphiteConfiguration);
+  }
+
+  @Test
+  public void unenabledGraphiteConfigurationShouldGetOneReporter() {
```
massdosage: unenabledGraphiteConfigurationShouldGetOneReporter -> disabledGraphiteConfiguration

("disabled" better than "unenabled" and also if you don't say what the expected behaviour is in the method name its one less thing to modify if the behaviour changes)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTest.java
```diff
@@ -0,0 +1,56 @@
...
+    monitoringConfiguration.init();
+    Set<Closeable> reporters = monitoringConfiguration.getReporters();
+    assertThat(reporters.size(), is(1));
+  }
+
+  @Test
+  public void enabledGraphiteConfigurationShouldGetTwoReporters() {
```
massdosage: enabledGraphiteConfigurationShouldGetTwoReporters -> enabledGraphiteConfiguration

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTest.java
```diff
@@ -0,0 +1,56 @@
...
+    monitoringConfiguration = new MonitoringConfiguration();
+    monitoringConfiguration.setMetricRegistry(metricRegistry);
+    monitoringConfiguration.setGraphiteConfiguration(graphiteConfiguration);
+  }
+
+  @Test
+  public void unenabledGraphiteConfigurationShouldGetOneReporter() {
+    monitoringConfiguration.init();
+    Set<Closeable> reporters = monitoringConfiguration.getReporters();
+    assertThat(reporters.size(), is(1));
```
massdosage: You could assert that the type of the reporter is JmxReporter

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTest.java
```diff
@@ -0,0 +1,56 @@
...
+  public void enabledGraphiteConfigurationShouldGetTwoReporters() {
+    when(graphiteConfiguration.isEnabled()).thenReturn(true);
+    when(graphiteConfiguration.getHost()).thenReturn("host");
+    when(graphiteConfiguration.getPort()).thenReturn(42);
+    when(graphiteConfiguration.getPollInterval()).thenReturn((long) 42);
+    when(graphiteConfiguration.getPollIntervalTimeUnit()).thenReturn(TimeUnit.NANOSECONDS);
+    monitoringConfiguration.setGraphiteConfiguration(graphiteConfiguration);
+    monitoringConfiguration.init();
+    Set<Closeable> reporters = monitoringConfiguration.getReporters();
+    assertThat(reporters.size(), is(2));
```
massdosage: Maybe a bit fiddly but you could assert that one of them is a JmxReporter and the other a GraphiteReporter

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/yaml/AdvancedPropertyUtilsTest.java
```diff
@@ -1,5 +1,5 @@
 /**
```
massdosage: This test looks good.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/tunneling/HiveMetaStoreClientSupplierTest.java
AnanaMJ: Actually, a mocked HiveConf works. I probably thought that because I need the metastore for the other tests in the tunnelling package, then I need it for this one too.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
AnanaMJ: When PrefixBasedDatabaseMappingService is constructed, if the list of metastore passed as parameters contains a primary metastore, a local variable, `primaryDatabaseMapping` is set to map to the primary, otherwise it is null. When `service.primaryDatabaseMapping()` is called, `primaryDatabaseMapping` is checked, and if it's null, an exception is thrown.
In this test, the prefix service doesn't have a primary passed through.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
AnanaMJ: Actually, after going through this test again, there is no need for that metastore. Only a list of non-primary metastores is needed, and a list with one element (federatedMetastore) is fine.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [TBD] - TBD
 ### Changed
 * Minimum supported Java version is now 8 (was 7). See [#108](https://github.com/HotelsDotCom/waggle-dance/issues/108).
+* Increased coverage for waggle-dance-core. See [#25](https://github.com/HotelsDotCom/waggle-dance/issues/25) and [#100](https://github.com/HotelsDotCom/waggle-dance/issues/100).
```
massdosage: We normally only put endiuser facing items in the CHANGELOG, most of them won't really care about changes made to unit tests so I'd remove this.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import com.codahale.metrics.JmxReporter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.graphite.GraphiteReporter;
+
+import com.hotels.bdp.waggledance.conf.GraphiteConfiguration;
+
+//@RunWith(MockitoJUnitRunner.class)
```
massdosage: Remove this commented out line

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/server/FederatedHMSHandlerFactoryTest.java
```diff
@@ -61,4 +61,13 @@ public void typical() throws Exception {
     assertThat(handler, is(instanceOf(FederatedHMSHandler.class)));
   }
 
+  @Test
+  public void prefixecDatabase() throws Exception {
```
massdosage: prefixecDatabase > prefixedDatabase

PR #112, opened 2018-07-26
Replace Cobertura with Jacoco

### LINKED ISSUE #109: Replace usage of Cobertura with Jacoco
As a developer
I want to use Jacoco for code coverage rather than Cobertura since the latter is pretty much a dead project (and doesn't handle Java 8 code)
So that I have a future proof code coverage solution.

**Acceptance Criteria**
- Usage of Cobertura in Travis replaced with Jacoco
- Coveralls reports checked to still work

massdosage:
Fixes #109

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ before_install:
 install:
   - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -P travis
 script:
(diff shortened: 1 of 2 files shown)


massdosage:
This also needs a change in the parent in order to work since Waggle Dance sets test.arguments to give Surefire more memory: https://github.com/HotelsDotCom/hotels-oss-parent/pull/21

Issue #115, opened 2018-07-30
Performance degradation on tunneled connection


bAndie91:
I've experienced performance degradation when upgraded from 2.2.2 to 2.3.7. 
see measurements in attachment which was made by Spark application calling `spark.catalog.listTables()`.
newer WD is 3 times slower impacting the ssh-tunneled connections (see highlighted rows) the most.
![image](https://user-images.githubusercontent.com/802732/43392117-f3039f6c-93f2-11e8-92f6-a2949869629e.png)

how much can it be eliminated?

rambrus:
@bAndie91 , @patduin: I did some investigation and found this:

**Test Case 1 - listTables on non-tunneled connection**

wd version | run-1 | run-2
-- | -- | --
2.3.0 | 20 s | 19 s
2.3.1 | 20 s | 23 s
2.3.2 | 25 s | 24 s
2.3.3 | 36 s | 32 s
2.3.4 | 28 s | 25 s
2.3.5 | 44 s | 56 s
2.3.6 | 48 s | 46 s

![image](https://user-images.githubusercontent.com/10614761/43522358-664fa04a-9599-11e8-8198-06e626fa6ead.png)

**Test Case 2 - listTables on tunneled connection**

wd version | run-1 | run-2
-- | -- | --
2.3.0 | 5 m 37 s | 4 m 52 s
2.3.1 | 4 m 59 s | 5 m 04 s
2.3.2 | 5 m 15 s | 5 m 11 s
2.3.3 | 7 m 27 s | 7 m 12 s
2.3.4 | 5 m 14 s | 5 m 15 s
2.3.5 | 13 m 02 s | 13 m 27 s
2.3.6 | 12 m 33 s | 13 m 07 s

![image](https://user-images.githubusercontent.com/10614761/43522404-8d9ad340-9599-11e8-94af-cd90e1025a52.png)

**Summary**
I had 2 runs for both test cases, the test durations are pretty much consistent between runs and we can observe ~150% performance degradation between 2.3.4 and 2.3.5 releases.

IMPORTANT: The performance degradation does not seem to be specific to *tunneled connection*, the same trend can be observed in both cases.

patduin:
Excellent work! 
Got one more questions in your WD configuration are all metastores reachable and responding or is one of them down?

rambrus:
@bAndie91 can you please answer the question above?

I also checked the spark logs if they can contain any unusual error.
[...]

PR #118, opened 2018-08-08
Issue 115, performance degradation addressed

patduin:
trying to fix(or at least improve) on #115 

I've changed the code around a bit to allow for the getStatus call to be done only for tunneled connections. I haven't found any other way to figure out how a tunneled connection is working or not. Jsch has an ensureOpen, but that works even if the metatstore is not reachable, they only way to find it is doing some kind of call. We should not have to do this for the direct connections so we won't hit the extra call performance hit.

### DIFF
--- a/waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
+++ b/waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
@@ -18,5 +18,4 @@ package com.hotels.bdp.waggledance.api.model;
 public enum ConnectionType {
   DIRECT_CONNECTION,
   TUNNELED;
-
 }
--- a/waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
+++ b/waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
@@ -1,0 +1,22 @@
+/**
+ * Copyright (C) 2016-2018 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.api.model;
+
+public enum ConnectionType {
+  DIRECT_CONNECTION,
+  TUNNELED;
+
+}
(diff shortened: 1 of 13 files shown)


waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.api.model;
+
+public enum ConnectionType {
+  DIRECT_CONNECTION,
+  TUNNELED;
+
```
teabot: remove line

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.api.model;
+
+public enum ConnectionType {
+  DIRECT_CONNECTION,
```
massdosage: I think I'd prefer it if this was just called `DIRECT` as we already know it's related to a connection from the Enum name.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/SimpleFederationStatusService.java
```diff
@@ -21,17 +21,17 @@
...
-import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIface;
-import com.hotels.bdp.waggledance.client.CloseableThriftHiveMetastoreIfaceClientFactory;
+import com.hotels.bdp.waggledance.mapping.model.MetaStoreMapping;
+import com.hotels.bdp.waggledance.mapping.service.MetaStoreMappingFactory;
 
 @Service
 public class SimpleFederationStatusService implements FederationStatusService {
 
-  private final CloseableThriftHiveMetastoreIfaceClientFactory metaStoreClientFactory;
+  private final MetaStoreMappingFactory metaStoreMappingFactory;
```
massdosage: This is better, nice.

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/ConnectionType.java
patduin: sure will change.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImplTest.java
```diff
@@ -47,10 +51,13 @@
...
 
   private MetaStoreMapping metaStoreMapping;
+  private MetaStoreMapping tunneledMetaStoreMapping;
 
   @Before
   public void init() {
-    metaStoreMapping = new MetaStoreMappingImpl(DATABASE_PREFIX, NAME, client, accessControlHandler);
+    metaStoreMapping = new MetaStoreMappingImpl(DATABASE_PREFIX, NAME, client, accessControlHandler, DIRECT);
+    tunneledMetaStoreMapping = new MetaStoreMappingImpl(DATABASE_PREFIX, NAME, client, accessControlHandler, TUNNELED);
+
```
massdosage: OMG extra blank line!

PR #125, opened 2018-08-15
reporting metrics for graphite before close, also removed systemExit

patduin:
Trying to fix our Travis builds

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -79,7 +79,9 @@ public class WaggleDance {
       }
       throw e;
     }
-    System.exit(exitCode);
+    if (exitCode != 0) {
+      throw new Exception("Waggle didn't exit properly see logs for errors, exitCode=" + exitCode);
+    }
   }
 
   private static SpringApplication registerListeners(SpringApplication application) {
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
@@ -80,7 +80,7 @@ public class WaggleDance {
       throw e;
     }
     if (exitCode != 0) {
-      throw new Exception("Waggle didn't exit properly see logs for errors, exitCode=" + exitCode);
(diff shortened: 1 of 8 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
```diff
@@ -79,7 +79,9 @@ public static void main(String[] args) throws Exception {
       }
       throw e;
     }
-    System.exit(exitCode);
+    if (exitCode != 0) {
+      throw new Exception("Waggle didn't exit properly see logs for errors, exitCode=" + exitCode);
```
massdosage: Waggle -> Waggle Dance

Won't this now change the exit code into whatever a thrown exception causes the exit code to be? Maybe we don't care as long as we're sure this isn't 0?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
```diff
@@ -79,7 +79,9 @@ public static void main(String[] args) throws Exception {
       }
       throw e;
     }
-    System.exit(exitCode);
+    if (exitCode != 0) {
```
massdosage: Probably also warrants an entry in the changelog just in case.

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test;
+
+import com.codahale.metrics.MetricRegistry;
+
+import com.hotels.bdp.waggledance.conf.GraphiteConfiguration;
+import com.hotels.bdp.waggledance.junit.ServerSocketRule;
+
+public class MonitoringConfigurationIntegrationTest {
+
+  public @Rule ServerSocketRule graphite = new ServerSocketRule();
```
patduin: ServerSocketRule is in this maven module so we had to put the test here as well to avoid a cyclical dependency

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+  private final GraphiteConfiguration graphiteConfiguration = new GraphiteConfiguration();
+  private final MetricRegistry metricRegistry = new MetricRegistry();
+  private MonitoringConfiguration monitoringConfiguration;
+
+  @Test
+  public void allMetricsAreLoggedWhenPollItervalIsHigh() throws Exception {
```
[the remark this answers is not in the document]
patduin: yes

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/WaggleDance.java
patduin: sorry

Issue #103, opened 2018-07-19
Refactor duplicated SSH code


massdosage:
_As a developer
I want code that is duplicated by code present in [hcommon-ssh](https://github.com/HotelsDotCom/hcommon-ssh/) removed from WaggleDance
So that I have less code to maintain, can fix bugs in one place etc._

**Acceptance Criteria**
The following are candidates for refactoring, further investigation might reveal that it's not feasible so for all of the below we should either make the change or document why we decided not to:
- Replace `com.hotels.bdp.waggledance.api.model.MetastoreTunnel` with `com.hotels.hcommon.ssh.SshSettings`
- If the above is successful create a similar ticket to this for [CircusTrain](https://github.com/HotelsDotCom/circus-train) to make the same change for `com.hotels.bdp.circustrain.api.conf.MetastoreTunnel`
- Replace `com.hotels.bdp.waggledance.api.validation.constraint.TunnelRoute` with `com.hotels.hcommon.ssh.validation.constraint.TunnelRoute` and
`com.hotels.bdp.waggledance.api.validation.validator.TunnelRouteValidator` with `com.hotels.hcommon.ssh.validation.validator.TunnelRouteValidator`
- If the above is successful create a similar ticket to this for [CircusTrain](https://github.com/HotelsDotCom/circus-train) to make the same change for `com.hotels.bdp.circustrain.api.validation.constraints.TunnelRoute` and `com.hotels.bdp.circustrain.api.validation.constraintvalidators.TunnelRouteValidator`)
- Look at `TunnelableFactorySupplier` and how we're using `WaggleDanceHiveConfVars`, could we replace some of this with `com.hotels.hcommon.ssh.SshSettings` or `com.hotels.hcommon.hive.metastore.client.tunnelling.TunnellingMetaStoreClientSupplierBuilder`? See https://github.com/HotelsDotCom/circus-train/commit/0272e36281bc70034c50d27a5822e160d80f3bc9 for a similar change to CircusTrain that could possibly be applied here.
- Check whether is any scope for refactoring/removing any code in `com.hotels.bdp.waggledance.client.TunnelingMetaStoreClientFactory`
- Above code tested out in our development environment in AWS via a tunnel to ensure tunnels still work correctly.

AnanaMJ:
[...]

PR #136, opened 2018-09-18
Refactor MetastoreTunnel to remove duplicated code

AnanaMJ:
Dependent on https://github.com/HotelsDotCom/hcommon-hive-metastore/pull/5, which would need to be released before this PR is merged.


AnanaMJ:
This is part of issue #103 about refactoring duplicated SSH code.

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnel.java
```diff
@@ -15,82 +15,6 @@
...
-
-  public String getStrictHostKeyChecking() {
-    return strictHostKeyChecking;
-  }
-
-  public void setStrictHostKeyChecking(String strictHostKeyChecking) {
-    this.strictHostKeyChecking = strictHostKeyChecking;
-  }
-}
+public class MetastoreTunnel extends SshSettingsMapper {}
```
massdosage: Nice, look at all that repeated code gone!

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
```diff
@@ -72,99 +66,77 @@ public void portTooHigh() {
...
     Set<ConstraintViolation<MetastoreTunnel>> violations = validator.validate(tunnel);
-
     assertThat(violations.size(), is(1));
   }
 
   @Test
   public void emptyKnownHosts() {
-    tunnel.setKnownHosts("");
-
+    tunnel.setKnownHosts(" ");
```
massdosage: This is now blank not empty, see the below test and compare.

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/WaggleDanceRunner.java
```diff
@@ -133,6 +139,25 @@ public Builder federate(String name, String remoteMetaStoreUris, AccessControlTy
       return this;
     }
 
+    public Builder federateWithMetastoreTunnel(
```
massdosage: Why do we need this? Is it used anywhere other than a test?

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnelTest.java
AnanaMJ: Good spot, thanks!

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/WaggleDanceRunner.java
AnanaMJ: It is just for the integration test, but I could not find any other way of making the integration test include something that takes metastore-tunnel options too.

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MetastoreTunnel.java
massdosage: Well, we don't want to mention the Metastore in hcommon-ssh as it's supposed to only know about ssh, not metastores. So the compromise was to keep this here so we have a descriptive and backwards compatible name for the YAML/Spring Boot properties but at least all the repeated code is gone.

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/WaggleDanceRunner.java
AnanaMJ: This is testing that the configuration can be loaded correctly. When I first tried to replace MetastoreTunnel with SshSettings, the bean couldn't be initialised. On a real environment, the change was failing, but the integration tests were all passing successfully. Where do you think it would be better to test this behaviour?
However, the test does pass when I run it on my IDE, but I wonder why...

pom.xml
```diff
@@ -8,7 +8,7 @@
   </parent>
 
   <artifactId>waggle-dance-parent</artifactId>
-  <version>3.0.1-SNAPSHOT</version>
+  <version>3.1.0-SNAPSHOT</version>
```
massdosage: You'll need to make this change in the pom files for all the sub-modules to. This should do the trick:

`mvn versions:set -DnewVersion=3.1.0-SNAPSHOT -DgenerateBackupPoms=false`

AnanaMJ: That didn't work, but I did it manually :)

massdosage: It should have, but maybe it failed since you'd already done the parent manually but not the others? Anyway, try it next time.

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/WaggleDanceRunner.java
patduin: discussed offline, but we decided to keep it as it is something that benefits from testing :)

massdosage: OK, good to know. I'm going to "resolve conversation" then!

waggle-dance-api/pom.xml
```diff
@@ -88,6 +88,13 @@
       <groupId>commons-lang</groupId>
       <artifactId>commons-lang</artifactId>
     </dependency>
+    
+    <!-- SSH library -->
+    <dependency>
+      <groupId>com.hotels</groupId>
+      <artifactId>hcommon-hive-metastore</artifactId>
```
patduin: ah we need this in core and api, better we make this a managed dependency in the parent pom and remove the version here and in the core/pom.xml

waggle-dance-core/pom.xml
```diff
@@ -14,7 +14,7 @@
     <mockito.version>2.8.47</mockito.version>
     <powermock.version>1.7.4</powermock.version>
     <jcabi-aspects.version>0.22.6</jcabi-aspects.version>
-    <hcommon-hive-metastore.version>1.1.0</hcommon-hive-metastore.version>
+    <hcommon-hive-metastore.version>1.2.0</hcommon-hive-metastore.version>
```
patduin: see my comment in the api/pom.xm, move this to the parent please

waggle-dance-core/pom.xml
```diff
@@ -187,6 +179,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+    
+    <!-- Hive common -->
```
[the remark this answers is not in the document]
massdosage: Well, it is actually Hive common as it contains a bunch of generic code to interact with the Metastore. It does also contains the code that allows one to communicate with a metastore via an SSH tunnel but the library is broader than just that.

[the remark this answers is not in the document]
abhimanyugupta07: We already have a <--hive--> section where we have Hive related Libraries added. IMHO, having it separately tagged as <--Hive-common--> is a bit confusing and doesn't give the reader context as to what purpose its being added for in the project. Is there any other use for this library in this project other than SSH tunnel ?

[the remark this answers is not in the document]
AnanaMJ: Yes, there is another use for this library. https://github.com/HotelsDotCom/circus-train/pull/76 is an example of what it can help with, and something similar to this will happen for WaggleDance too.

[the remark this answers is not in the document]
abhimanyugupta07: Alright. I will resolve this.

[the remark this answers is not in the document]
massdosage: I see what you're saying about "Hive common" being confusing. I think we should just remove the comment and assume that the name of the artifact is sufficient documentation ;) The "Hive" section was probably intended to group all the Apache Hive artifacts, not our own Hive-related ones.

PR #138, opened 2018-09-28
Feature/prefix readme clarification

massdosage:
See https://github.com/HotelsDotCom/waggle-dance/issues/134#issuecomment-422411943

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,7 +339,12 @@ NOTE: in the case of manual database resolution the configuration still requires
 
 ##### Database resolution: `PREFIXED`
 
-Waggle Dance can be configured to use a prefix when resolving the names of databases in its primary or federated metastores. All queries issued to Waggle Dance need to use fully qualified database names and the database names need to use the same prefixes configured here.
+Waggle Dance can be configured to use a prefix when resolving the names of databases in its primary or federated metastores. All queries issued to Waggle Dance need to use 
+fully qualified database names and the database names need to use the same prefixes configured here. In the example below Waggle Dance is configured to have a federated 
+metastore with the prefix `waggle_prod`. Because of this it will inspect the database names in all requests, and if they start with this configured prefix, it will route 
+the request to the configured matching metastore. The prefix will be removed for those requests as the underlying metastore knows nothing of the prefixes. So, the 
+query: `select * from waggle_prod_etldata.my_table` will effectively be translated into this query: `select * from etldata.my_table` on the federated metastore. If a 
+database is encountered that is not prefixed the primary metastore is used to resolve the database name.
 
 `waggle-dance-server.yml`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -351,14 +356,16 @@ Waggle Dance can be configured to use a prefix when resolving the names of datab
...
-
-Newly created databases are immediately accessible - no service restart is necessary.
+Note: When choosing a prefix ensure that it does not match the start of _any_ existing database names in your metastores. To illustrate the problem this would cause, 
+imagine you have a database in Metastore "A" named "my_database" and you configure another Metastore "B" with the prefix `my_`. Waggle Dance will register the prefix 
+and any requests for a database starting with `my_` will be routed to Metastore "B" even if they were intended to go to Metastore "A".
 
-Alternatively, Waggle Dance can be configured to use a static list of unprefixed databases in the configuration `waggle-dance-federations.yml`:`federated-meta-stores[n].mapped-databases`. Example configuration:
+In this mode any databases that are created while Waggle Dance is running will be automatically visible and will need to adhere to the naming rules described above 
+(e.g. not clash with the prefix). Alternatively, Waggle Dance can be configured to use a static list of unprefixed databases in the configuration 
+`waggle-dance-federations.yml`:`federated-meta-stores[n].mapped-databases`. Example configuration:
```
massdosage: @patduin if you look at line 380 below, is this correct? There is a name but no prefix, doesn't there still need to be a prefix for this to work? Maybe this is left over from when we allowed the name to be used as a prefix but decided to separate the two concepts?

AnanaMJ: I've just tried it and it's correct. If there are no clashing database names, it works fine, but with name clashes, an error is thrown.

massdosage: Having looked at it, I think the current config examples are missing the `prefix` so I have added them.

README.md
```diff
@@ -356,7 +356,8 @@ database is encountered that is not prefixed the primary metastore is used to re
       name: primary
       remote-meta-store-uris: thrift://primaryLocalMetastore:9083
     federated-meta-stores:
-      - name: waggle_prod_
+      - name: waggle_prod
+        prefix: waggle_prod
```
patduin: this is not necessary, by default the prefix is made up from the `name`+`_`. So `waggle_prod` becomes `waggle_prod_`

massdosage: Yeah, so this is what I changed, I added the `prefix` as the documentation refers to the `prefix` not the `name` so it was confusing. So it sounds like you can use either but the documentation isn't very clear on that. We should probably decide which one we prefer and then go through this whole document and make sure its consistent. I thought we separated `prefix` and `name` so that the one would show up in log files etc (a readable `name`) and the other would be used for prefixing which is clearer IMHO, but I can't remember all the details now.

massdosage:
OK @patduin I've had another go at further clarifying the prefix based on our discussion this morning. Could you please take another look?

PR #139, opened 2018-10-09
Refactor general metastore tunnelling code

### LINKED ISSUE #103: Refactor duplicated SSH code

AnanaMJ:
Fixes #103 
This change makes use of hcommon-hive-metastore and hcommon-ssh libraries.

Related PRs:
* https://github.com/HotelsDotCom/hcommon-ssh/pull/5
* https://github.com/HotelsDotCom/hcommon-hive-metastore/pull/7


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
```diff
@@ -38,21 +44,41 @@ public CloseableThriftHiveMetastoreIfaceClientFactory(MetaStoreClientFactory met
...
+      metaStoreClientFactory.setLocalhost(metastoreTunnel.getLocalhost());
     }
-    HiveConfFactory confFactory = new HiveConfFactory(Collections.<String> emptyList(), properties);
+
     return metaStoreClientFactory.newInstance(confFactory.newInstance(), "waggledance-" + name, 3);
   }
+
+  private SshSettings buildSshSettings(MetastoreTunnel metastoreTunnel) {
+    boolean strictHostKeyChecking = true;
+    if (metastoreTunnel.getStrictHostKeyChecking().toLowerCase() == "no") {
```
abhimanyugupta07: why not use String equals for comparision ?

pom.xml
```diff
@@ -43,7 +43,7 @@
     <beeju.version>1.2.1</beeju.version>
     <guava.version>23.0</guava.version>
     <guice.version>4.0</guice.version>
-    <hcommon-hive-metastore.version>1.2.0</hcommon-hive-metastore.version>
+    <hcommon-hive-metastore.version>1.2.1-SNAPSHOT</hcommon-hive-metastore.version>
```
abhimanyugupta07: Just putting a comment here to update these versions before merging the PR.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
```diff
@@ -41,55 +38,65 @@
...
-      LocalHiveConfFactory localHiveConfFactory,
       HiveMetaStoreClientSupplierFactory hiveMetaStoreClientSupplierFactory) {
-    this.tunnelableFactorySupplier = tunnelableFactorySupplier;
     this.defaultFactory = defaultFactory;
-    this.localHiveConfFactory = localHiveConfFactory;
     this.hiveMetaStoreClientSupplierFactory = hiveMetaStoreClientSupplierFactory;
   }
 
+  @Override
+  public void setTunnelableFactory(TunnelableFactory<CloseableThriftHiveMetastoreIface> tunnelableFactory) {
```
massdosage: So before these changes the constructor took this and localHost as paramaters and once you had constructed the factory then calles to `newInstance()` would always return the same type. Now theoretically you could construct the factory without setting this or the next one, make some calls to `newInstance()` and get a certain type of `CloseableThriftHiveMetastoreIface` back, then set these, call `newInstance()` again and get a different (tunnelled) type back. I see from the flow of code in the classes above that this wouldn't actually happen in how we use this factory but it is a bit unusual. How hard is it to get these two handed over via a constructor?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactoryTest.java
```diff
@@ -51,46 +66,61 @@ public void setUp() {
...
+  private void checkSshSettingsParameters(SshSettings sshSettings) {
+    assertThat(sshSettings.getLocalhost(), is(localhost));
+    assertThat(sshSettings.getSshPort(), is(port));
+    assertThat(sshSettings.getRoute(), is(route));
+    assertThat(sshSettings.getKnownHosts(), is(knownHosts));
+    assertThat(sshSettings.getPrivateKeys(), is(Collections.singletonList(privateKeys)));
+    assertThat(sshSettings.getSessionTimeout(), is(timeout));
+  }
+
+  private MetastoreTunnel setMetastoreTunnel() {
```
massdosage: `createMetastoreTunnel()` is probably a better name for this.

AnanaMJ:
I haven't changed the changelog because it has this line right now:
> Refactored general metastore tunnelling code to leverage hcommon-hive-metastore libraries. See #103.

That is from the previous PR that was related to removing MetastoreTunneling and other classes that it was using, but it's relevant for the changes that were done this time because it is part of the same ticket. The minor version was changed to 3.1.0-SNAPSHOT as a result of that change, so the version that will be released is 3.1.0. I don't need to change it, right?

I could, however, add a line saying that WaggleDanceHiveConfVars was removed. Do you think that's a good idea, @massdosage ?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
```diff
@@ -38,21 +44,41 @@ public CloseableThriftHiveMetastoreIfaceClientFactory(MetaStoreClientFactory met
...
-      properties.put(WaggleDanceHiveConfVars.SSH_ROUTE.varname, metastoreTunnel.getRoute());
-      properties.put(WaggleDanceHiveConfVars.SSH_KNOWN_HOSTS.varname, metastoreTunnel.getKnownHosts());
-      properties.put(WaggleDanceHiveConfVars.SSH_PRIVATE_KEYS.varname, metastoreTunnel.getPrivateKeys());
-      properties.put(WaggleDanceHiveConfVars.SSH_SESSION_TIMEOUT.varname, String.valueOf(metastoreTunnel.getTimeout()));
-      properties
-          .put(WaggleDanceHiveConfVars.SSH_STRICT_HOST_KEY_CHECKING.varname,
-              metastoreTunnel.getStrictHostKeyChecking());
+
+      sshSettings = buildSshSettings(metastoreTunnel);
+      metaStoreClientFactory.setTunnelableFactory(new TunnelableFactory<>(sshSettings));
```
patduin: These settters should go, `metaStoreClientFactory` should not be changed by this class, what if the factory instance is injected somewhere else we'll have very strange sideeffect.

massdosage:
> I haven't changed the changelog because it has this line right now:
> 
> > Refactored general metastore tunnelling code to leverage hcommon-hive-metastore libraries. See #103.
> 
> That is from the previous PR that was related to removing MetastoreTunneling and other classes that it was using, but it's relevant for the changes that were done this time because it is part of the same ticket. The minor version was changed to 3.1.0-SNAPSHOT as a result of that change, so the version that will be released is 3.1.0. I don't need to change it, right?
> 
> I could, however, add a line saying that WaggleDanceHiveConfVars was removed. Do you think that's a good idea, @massdosage ?

Ah OK, no, I think what is there is fine then. Nobody should be using WaggleDance as a library so `WaggleDanceHiveConfVars` isn't something they should even be aware of really.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
AnanaMJ: I don't think it should be hard to do it, but I'm going to try it out and update you on this.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
AnanaMJ: Ok, I'll try to pass it directly to TunnelingMetaStoreClientFactory and update the PR once I have something that works.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
```diff
@@ -41,55 +38,65 @@
   @VisibleForTesting
   final MethodChecker METHOD_CHECKER = new MetastoreClientMethodChecker();
 
-  private final TunnelableFactorySupplier tunnelableFactorySupplier;
+  private HiveConf localHiveConf;
```
patduin: Don't store state in Factories. remove please keep that scoped to the `newInstance`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
```diff
@@ -41,55 +38,65 @@
...
+  private final TunnelableFactory<CloseableThriftHiveMetastoreIface> tunnelableFactory;
+  private final String localhost;
   private final MetaStoreClientFactory defaultFactory;
-  private final LocalHiveConfFactory localHiveConfFactory;
   private final HiveMetaStoreClientSupplierFactory hiveMetaStoreClientSupplierFactory;
 
   public TunnelingMetaStoreClientFactory() {
-    this(new TunnelableFactorySupplier(), new DefaultMetaStoreClientFactory(), new LocalHiveConfFactory(),
-        new HiveMetaStoreClientSupplierFactory());
+    this(new DefaultMetaStoreClientFactory(), new HiveMetaStoreClientSupplierFactory(), null, null);
```
patduin: we use a `DefaultMetaStoreClientFactory` to fall back on, but we also do that in the helper, we should probably remove it here so this only deals with tunnels

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -52,7 +53,8 @@ public MetaStoreMappingFactoryImpl(
 
   private CloseableThriftHiveMetastoreIface createClient(AbstractMetaStore metaStore) {
     try {
-      return metaStoreClientFactory.newInstance(metaStore);
+      MetastoreClientFactoryHelper helper = new MetastoreClientFactoryHelper(metaStore);
```
patduin: I find it really weird that I have to create two object with the same argument `metastore`. Sure the factory can create it as well? If this is done to make it more testable then the problem is just shifted right and this method now is just as hard to test.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactoryTest.java
```diff
@@ -15,82 +15,61 @@
...
-
-    HiveConf hiveConf = hiveConfCaptor.getValue();
-    assertThat(hiveConf.getVar(ConfVars.METASTOREURIS), is(THRIFT_URI));
-    assertThat(hiveConf.get(WaggleDanceHiveConfVars.SSH_LOCALHOST.varname), is("local-machine"));
-    assertThat(hiveConf.get(WaggleDanceHiveConfVars.SSH_PORT.varname), is("2222"));
-    assertThat(hiveConf.get(WaggleDanceHiveConfVars.SSH_ROUTE.varname), is("a -> b -> c"));
-    assertThat(hiveConf.get(WaggleDanceHiveConfVars.SSH_KNOWN_HOSTS.varname), is("knownHosts"));
-    assertThat(hiveConf.get(WaggleDanceHiveConfVars.SSH_PRIVATE_KEYS.varname), is("privateKeys"));
-    assertThat(hiveConf.get(WaggleDanceHiveConfVars.SSH_SESSION_TIMEOUT.varname), is("123"));
+    when(helper.get()).thenReturn(tunnelingMetaStoreClientFactory);
```
patduin: The class doesn't know about tunnels anymore nor should the test, you are just testing mocks

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/MetastoreClientFactoryHelper.java
```diff
@@ -0,0 +1,61 @@
...
+      SshSettings sshSettings = buildSshSettings(metastoreTunnel);
+      metaStoreClientFactory = new TunnelingMetaStoreClientFactory(new TunnelableFactory<>(sshSettings),
+          metastoreTunnel.getLocalhost());
+    }
+    return metaStoreClientFactory;
+  }
+
+  private SshSettings buildSshSettings(MetastoreTunnel metastoreTunnel) {
+    boolean strictHostKeyChecking = true;
+    if (metastoreTunnel.getStrictHostKeyChecking().toLowerCase().equals("no")) {
```
patduin: this logic maybe should go to the MetastoreTunnel? add a `Boolean isStrictHostKeyChecking()` or the SShSetting should handle it.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
AnanaMJ: So instead of creating a new `DefaultMetaStoreClientFactory` when there is no `tunnelableFactory` (line 71), I could just return an exception, right?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactoryTest.java
AnanaMJ: Should I just remove this test then?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/MetastoreClientFactoryHelper.java
AnanaMJ: That's a great idea, I'll add that 👍

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
AnanaMJ: If I remove this, then I won't be able to test if the local hive conf gets the correct parameters.
Apparently there was an issue before where the code was changed, and instead of getting local port and local host, `localHiveConf` got other parameters. Because that happened before, I thought it would be a good idea to have a test that prevents that from happening again.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
AnanaMJ: So instead of creating a new `DefaultMetaStoreClientFactory` when there is no `tunnelableFactory` (line 71), I could just return an exception, right?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/tunnelling/TunnelingMetaStoreClientFactory.java
AnanaMJ: If I remove this, then I won't be able to test if the local hive conf gets the correct parameters.
Apparently there was an issue before where the code was changed, and instead of getting local port and local host, `localHiveConf` got other parameters. Because that happened before, I thought it would be a good idea to have a test that prevents that from happening again.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
```diff
@@ -15,44 +15,32 @@
...
 
-import com.hotels.bdp.waggledance.api.model.AbstractMetaStore;
-import com.hotels.hcommon.hive.metastore.client.tunnelling.MetastoreTunnel;
 import com.hotels.hcommon.hive.metastore.conf.HiveConfFactory;
+import com.hotels.hcommon.hive.metastore.util.MetaStoreUriNormaliser;
 
 public class CloseableThriftHiveMetastoreIfaceClientFactory {
 
-  private final MetaStoreClientFactory metaStoreClientFactory;
+  public CloseableThriftHiveMetastoreIface newInstance(MetastoreClientFactorySupplier helper) {
```
patduin: rename `helper-> supplier`

AnanaMJ:
Replaced by https://github.com/HotelsDotCom/waggle-dance/pull/143 for a better approach.

PR #146, opened 2018-11-09
synchronous socket is deterministic test

patduin:
@massdosage reported tests still failing sporadicly, hopefully this really gets rid of it

### DIFF
--- a/waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/junit/ServerSocketRule.java
+++ b/waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/junit/ServerSocketRule.java
@@ -19,16 +19,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
-import java.nio.channels.AsynchronousServerSocketChannel;
-import java.nio.channels.AsynchronousSocketChannel;
 import java.nio.channels.Channels;
-import java.nio.channels.CompletionHandler;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
+import java.nio.channels.ServerSocketChannel;
+import java.nio.channels.SocketChannel;
 
 import org.junit.rules.ExternalResource;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/junit/ServerSocketRule.java
```diff
@@ -45,104 +38,49 @@
...
-        requestCount--;
-      } catch (Exception e) {
-        throw new RuntimeException("Error while waiting for request completion", e);
+      SocketChannel socketChannel = serverSocketChannel.accept();
+      while (socketChannel != null) {
+        requests++;
+        try (InputStream input = Channels.newInputStream(socketChannel)) {
+          ByteStreams.copy(input, output);
+        }
+        socketChannel = serverSocketChannel.accept();
```
massdosage: Why do you need to do this?

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/junit/ServerSocketRule.java
```diff
@@ -45,104 +38,49 @@
...
-      address = (InetSocketAddress) listener.getLocalAddress();
+      serverSocketChannel = (ServerSocketChannel) ServerSocketChannel
+          .open()
+          .bind(new InetSocketAddress(0))
+          .configureBlocking(false);
+      address = (InetSocketAddress) serverSocketChannel.getLocalAddress();
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
   }
```
massdosage: With respect to all the code that's gone, any idea if it served much purpose? Are there new edge cases we might be exposed to? It seemed to be handling cases where you'd want to wait for input that had been previously been sent etc. Not sure we need it all, just curious what you thought after looking into in more detail. This looks much more straightforward and doesn't seem to exhibit the issue with reads returning data out of order that we've been seeing.

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/junit/ServerSocketRule.java
patduin: an `accept()` for every request, see the `expectMultipleRequests` test it loops through this 3 times

waggle-dance-integration-tests/src/main/java/com/hotels/bdp/waggledance/junit/ServerSocketRule.java
patduin: hard to tell, there were some bugs on the sending side (where we didn't flush explicitly) so some of them were added to fix that until we fixed the flushing. 
But I am not sure of the whole history. This doesn't seem to block, requests are pending until you `accept` at which point you can read it. It's probably less performant this way but more deterministic which is fine cause we use this is meant for tests.

PR #149, opened 2019-01-04
Checkstyle and findbugs refactoring

AnanaMJ:
Checkstyle warnings: 320 -> 9
Findbugs warnings: 26 -> 9

Remaining checkstyle and findbugs warnings:
* Checkstyle
  * [waggle-dance-core-checkstyle.html.zip](https://github.com/HotelsDotCom/waggle-dance/files/2732632/waggle-dance-core-checkstyle.html.zip)
  * [waggle-dance-integration-checkstyle.html.zip](https://github.com/HotelsDotCom/waggle-dance/files/2727936/waggle-dance-integration-checkstyle.html.zip)
* Findbugs
  * [waggle-dance-core-findbugs.html.zip](https://github.com/HotelsDotCom/waggle-dance/files/2727937/waggle-dance-core-findbugs.html.zip)
  * [waggle-dance-integration-findbugs.html.zip](https://github.com/HotelsDotCom/waggle-dance/files/2727939/waggle-dance-integration-findbugs.html.zip)

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
@@ -50,7 +50,7 @@ public class CloseableThriftHiveMetastoreIfaceClientFactory {
     }
     Map<String, String> properties = new HashMap<>();
     properties.put(ConfVars.METASTOREURIS.varname, uris);
-    HiveConfFactory confFactory = new HiveConfFactory(Collections.<String> emptyList(), properties);
+    HiveConfFactory confFactory = new HiveConfFactory(Collections.<String>emptyList(), properties);
     return defaultMetaStoreClientFactory
         .newInstance(confFactory.newInstance(), "waggledance-" + name, DEFAULT_CLIENT_FACTORY_RECONNECTION_RETRY);
   }
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
@@ -19,6 +19,7 @@ import static com.hotels.bdp.waggledance.api.model.ConnectionType.TUNNELED;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
@@ -43,7 +44,7 @@ public class CloseableThriftHiveMetastoreIfaceClientFactory {
 
   public CloseableThriftHiveMetastoreIface newInstance(AbstractMetaStore metaStore) {
     String uris = MetaStoreUriNormaliser.normaliseMetaStoreUris(metaStore.getRemoteMetaStoreUris());
-    String name = metaStore.getName().toLowerCase();
+    String name = metaStore.getName().toLowerCase(Locale.ROOT);
     if (metaStore.getConnectionType() == TUNNELED) {
       return tunnelingMetaStoreClientFactory
           .newInstance(uris, metaStore.getMetastoreTunnel(), name, DEFAULT_CLIENT_FACTORY_RECONNECTION_RETRY);
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2018 Expedia Inc.
+ * Copyright (C) 2016-2019 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,7 +32,7 @@ import com.hotels.hcommon.hive.metastore.exception.MetastoreUnavailableException
 
 public class DefaultMetaStoreClientFactory implements MetaStoreClientFactory {
 
-  static Class<?>[] INTERFACES = new Class<?>[] { CloseableThriftHiveMetastoreIface.class };
+  static final Class<?>[] INTERFACES = new Class<?>[] { CloseableThriftHiveMetastoreIface.class };
 
   private static class ReconnectingMetastoreClientInvocationHandler implements InvocationHandler {
     private static final Logger LOG = LoggerFactory.getLogger(ReconnectingMetastoreClientInvocationHandler.class);
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
@@ -34,7 +34,7 @@ public class DefaultMetaStoreClientFactory implements MetaStoreClientFactory {
 
   static final Class<?>[] INTERFACES = new Class<?>[] { CloseableThriftHiveMetastoreIface.class };
 
-  private static class ReconnectingMetastoreClientInvocationHandler implements InvocationHandler {
+  private static final class ReconnectingMetastoreClientInvocationHandler implements InvocationHandler {
     private static final Logger LOG = LoggerFactory.getLogger(ReconnectingMetastoreClientInvocationHandler.class);
 
     private final ThriftMetastoreClientManager base;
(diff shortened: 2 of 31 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
```diff
@@ -43,14 +44,14 @@ public CloseableThriftHiveMetastoreIfaceClientFactory(
 
   public CloseableThriftHiveMetastoreIface newInstance(AbstractMetaStore metaStore) {
     String uris = MetaStoreUriNormaliser.normaliseMetaStoreUris(metaStore.getRemoteMetaStoreUris());
-    String name = metaStore.getName().toLowerCase();
+    String name = metaStore.getName().toLowerCase(Locale.ROOT);
```
AnanaMJ: Conversions to upper or lowercase use the platform's default encoding, which may result in improper conversions when used with international characters. The recommended way is:
* String.toUpperCase( Locale l )
* String.toLowerCase( Locale l )

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
```diff
@@ -53,7 +53,7 @@
   private ThriftHiveMetastore.Iface client = null;
   private TTransport transport = null;
   private boolean isConnected = false;
-  private URI metastoreUris[];
+  private URI[] metastoreUris;
```
AnanaMJ: There are multiple styles of array type definitions in Java: 
* Java style: public static void main(String[] args)
* C style: public static void main(String args[])

The Checkstyle configuration opts for the first one by default.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -32,7 +32,7 @@
 
 public class DefaultMetaStoreClientFactory implements MetaStoreClientFactory {
 
-  static final Class<?>[] INTERFACES = new Class<?>[] { CloseableThriftHiveMetastoreIface.class };
+  static Class<?>[] INTERFACES = new Class<?>[] { CloseableThriftHiveMetastoreIface.class };
```
AnanaMJ: This is a mistake; I'll undo with the next commit.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -181,7 +182,9 @@ private boolean includeInResults(MetaStoreMapping metaStoreMapping, String prefi
   @Override
   public DatabaseMapping databaseMapping(@NotNull String databaseName) {
     // Find a Metastore with a prefix
-    for (String metastorePrefix : mappingsByPrefix.keySet()) {
+    Iterator<Entry<String, DatabaseMapping>> iterator = mappingsByPrefix.entrySet().iterator();
+    while (iterator.hasNext()) {
+      String metastorePrefix = iterator.next().getKey();
```
abhimanyugupta07: From the warning that you sent me for this change:

> makes inefficient use of keySet iterator instead of entrySet iterator
This method accesses the value of a Map entry, using a key that was retrieved from a keySet iterator. It is more efficient to use an iterator on the entrySet of the map, to avoid the Map.get(key) lookup.

So, to complete this optimization, you need to use entry.getValue() on line 189.

`Entry<String, DatabaseMapping> entry = iteratory.next();`
`DatabaseMapping databaseMapping = entry.getValue();`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/compatibility/HiveCompatibleThriftHiveMetastoreIfaceFactory.java
```diff
@@ -31,7 +31,7 @@
 
   private static final Logger log = LoggerFactory.getLogger(HiveCompatibleThriftHiveMetastoreIfaceFactory.class);
 
-  class ThriftMetaStoreClientInvocationHandler implements InvocationHandler {
+  private static class ThriftMetaStoreClientInvocationHandler implements InvocationHandler {
```
AnanaMJ: The warning is:
```
Should com.hotels.bdp.waggledance.client.compatibility.HiveCompatibleThriftHiveMetastoreIfaceFactory$ThriftMetaStore
ClientInvocationHandler be a _static_ inner class?

This class is an inner class, but does not use its embedded reference to the object which created it.  This reference 
makes the instances of the class larger, and may keep the reference to the creator object alive longer than 
necessary.  If possible, the class should be made static.
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/compatibility/HiveCompatibleThriftHiveMetastoreIfaceFactory.java
```diff
@@ -71,7 +71,7 @@ private Object invokeCompatibility(Method method, Object[] args) throws Throwabl
 
     private Class<?>[] getTypes(Object[] args) {
       if (args == null) {
-        return (Class<?>[]) args;
+        return new Class<?>[] {};
```
AnanaMJ: `args` is already null, so this is equivalent, but doesn't load a null variable.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -192,28 +192,28 @@ public void shutdown() {
 
   //////////////////////////////
 
-  @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
   @Override
+  @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
```
AnanaMJ: If the logger annotation is above the override one, checkstyle doesn't know it's an overriden method

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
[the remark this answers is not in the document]
massdosage: Was the error here that an array isn't immutable so therefore shouldn't be a constant with all upper case? If so this could just be named `interfaces` rather than `INTERFACES` and then the `final` doesn't really add much. If we really don't want the contents changed then consider using some kind of immutable array or other collection but I'm OK with just renaming.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
massdosage: I'm not sure how this version ever slipped in!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/compatibility/HiveCompatibleThriftHiveMetastoreIfaceFactory.java
massdosage: Yeah, this is clearer.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
massdosage: Yes, good spot.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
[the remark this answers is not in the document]
AnanaMJ: I looked through the warnings on Jenkins and there's no warning about this one. I must have removed the "final" by mistake. Should I rename to `interfaces` anyway?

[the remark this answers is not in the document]
massdosage: I think just put the final back as the intention is for it to be a constant even if it technically isn't.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -181,9 +182,12 @@ private boolean includeInResults(MetaStoreMapping metaStoreMapping, String prefi
   @Override
   public DatabaseMapping databaseMapping(@NotNull String databaseName) {
     // Find a Metastore with a prefix
-    for (String metastorePrefix : mappingsByPrefix.keySet()) {
+    Iterator<Entry<String, DatabaseMapping>> iterator = mappingsByPrefix.entrySet().iterator();
```
patduin: I find the left side much easier to read imo

AnanaMJ: I think the way it was before is more readable, so I think this is mainly a question of whether we'd rather change the code to eliminate the warnings or silence the warnings because we don't care about them.

massdosage: It is more readable but less efficient, not that the efficiency gains in this case will make any real difference. I  do think that the general pattern of iterating over the entries rather than the keys when you use both the keys and values is preferable though.

PR #151, opened 2019-01-14
Support regex for MANUAL database resolution

### LINKED ISSUE #147: mapped-databases should support regex for MANUAL database resolution
_as a user of `MANUAL` database resolution_
_I want to use regex in the `mapped-databases` property_
_So that I can simplify the configuration_

Note: The current README reads as if this is possible this is due we overloaded the `mapped-databases`, property and gave it a slightly different meaning depending on `PREFIXED` or `MANUAL` resolution.

Let's fix this by supporting regex in `MANUAL` resolution.

Acceptance Criteria:
* `mapped-databases` can handle regex values for `MANUAL` database resolution.

AnanaMJ:
Fixes #147

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
@@ -111,7 +111,6 @@ public class StaticDatabaseMappingService implements MappingEventListener {
       if (metaStoreMapping.isAvailable()) {
         mappableDatabases = getDatabasesFromPattern(metaStoreMapping.getClient(),
             federatedMetaStore.getMappedDatabases());
-        LOG.info("Added {} to the list of federated mappable databases", mappableDatabases.toString());
       }
       validateFederatedMetastoreDatabases(mappableDatabases, metaStoreMapping);
       DatabaseMapping databaseMapping = createDatabaseMapping(metaStoreMapping);
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
@@ -168,23 +168,12 @@ public class StaticDatabaseMappingService implements MappingEventListener {
 
   private List<String> getDatabasesFromPattern(List<String> allDatabases, List<String> mappedDatabases) {
     List<String> matchedDatabases = new ArrayList<String>();
-    // for (String mappedDatabase : mappedDatabases) {
-    // try {
     Whitelist whitelist = new Whitelist(mappedDatabases);
     for (String database : allDatabases) {
       if (whitelist.contains(database)) {
         matchedDatabases.add(database);
       }
     }
-    //
-    // Iterable<String> matches = Iterables.filter(allDatabases, Predicates.contains(Pattern.compile(mappedDatabase)));
-    // for (String matched : matches) {
-    // matchedDatabases.add(matched);
-    // }
-    // } catch (PatternSyntaxException e) {
-    // LOG.error("Could not match databases for '{}'", mappedDatabase, e);
-    // }
-    // }
     return matchedDatabases;
   }
 
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
@@ -28,8 +28,6 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
 
 import javax.validation.constraints.NotNull;
 
@@ -39,11 +37,9 @@ import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Predicates;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
@@ -58,6 +54,7 @@ import com.hotels.bdp.waggledance.mapping.service.MappingEventListener;
 import com.hotels.bdp.waggledance.mapping.service.MetaStoreMappingFactory;
 import com.hotels.bdp.waggledance.mapping.service.PanopticOperationHandler;
 import com.hotels.bdp.waggledance.server.NoPrimaryMetastoreException;
+import com.hotels.bdp.waggledance.util.Whitelist;
 
 public class StaticDatabaseMappingService implements MappingEventListener {
   private static final Logger LOG = LoggerFactory.getLogger(StaticDatabaseMappingService.class);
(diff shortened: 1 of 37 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -155,6 +162,18 @@ private void validateFederatedMetastoreDatabases(List<String> mappableDatabases,
     }
   }
 
+  private List<String> getDatabasesFromPattern(Iface client, List<String> mappedDatabases) {
+    List<String> matchedDatabases = new ArrayList<String>();
+    for (String mappedDatabase : mappedDatabases) {
+      try {
+        matchedDatabases.addAll(client.get_databases(mappedDatabase));
+      } catch (TException e) {
+        LOG.error("Could not match databases for '{}'", mappedDatabase, e);
```
AnanaMJ: This, for now, logs an error if the pattern/database name doesn't match with anything in the metastore. Should it actually stop the execution if the database regex/name is incorrect and no database is found?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingServiceTest.java
```diff
@@ -65,22 +65,26 @@
...
   public void init() throws Exception {
-    federatedMetastore.setMappedDatabases(Lists.newArrayList("federated_DB"));
+    federatedMetastore.setMappedDatabases(mappedFederatedDatabases);
 
     metaStoreMappingPrimary = mockNewMapping(true, primaryMetastore);
     when(metaStoreMappingPrimary.getClient()).thenReturn(primaryDatabaseClient);
     when(primaryDatabaseClient.get_all_databases()).thenReturn(Lists.newArrayList("primary_db"));
     metaStoreMappingFederated = mockNewMapping(true, federatedMetastore);
+    when(metaStoreMappingFederated.getClient()).thenReturn(federatedDatabaseClient);
+    when(federatedDatabaseClient.get_databases("federated_DB")).thenReturn(mappedFederatedDatabases);
```
AnanaMJ: This test class had to be changed quite a bit because there were many NPEs saying that there is no client found, or failed tests because the client wasn't returning anything for `get_databases`.

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -671,4 +672,24 @@ public void metastoreTunnelConfiguration() throws Exception {
     assertThat(metastoreTunnel.getPrivateKeys(), is(privateKeys));
   }
 
+  @Test
+  public void getDatabaseFromPatternManual() throws Exception {
```
AnanaMJ: Should I also add a test for getting database from pattern using PREFIXED database resolution?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
patduin: No if the pattern doesn't match we just don't include anything. For instance consider someone dropping a database that was configured before. We don't want WD to fail if that happens.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -155,6 +162,18 @@ private void validateFederatedMetastoreDatabases(List<String> mappableDatabases,
     }
   }
 
+  private List<String> getDatabasesFromPattern(Iface client, List<String> mappedDatabases) {
+    List<String> matchedDatabases = new ArrayList<String>();
+    for (String mappedDatabase : mappedDatabases) {
+      try {
+        matchedDatabases.addAll(client.get_databases(mappedDatabase));
```
patduin: I think the logic here needs to change. 
* We're doing a fetch for every "pattern" which is not very efficient.
* Some pattern might return duplicates on subsequent fetches
* The pattern allowed by the `client.get_database` call is I think something like: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-ShowDatabases. So almost regex but not completely. 
I think it is better if we use java regex to compile the patterns. I think we do this in other places already. Then we can do a `client.get_all_databases` and match the results with our patterns. That will be only one client call.
I think the Integration test succeeds cause the pattern matches what hive supports a more complex regex will probably fail.

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
patduin: that would be great!

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingServiceTest.java
```diff
@@ -95,6 +99,24 @@ private MetaStoreMapping mockNewMapping(boolean isAvailable, AbstractMetaStore m
...
+      List<String> mappedDatabases,
+      boolean availableMapping) {
+    FederatedMetaStore newMetastore = newFederatedInstance(name, uri);
+    newMetastore.setMappedDatabases(mappedDatabases);
+    MetaStoreMapping newMapping = mockNewMapping(availableMapping, newMetastore);
+    when(metaStoreMappingFactory.newInstance(newMetastore)).thenReturn(newMapping);
+    when(newMapping.getClient()).thenReturn(federatedDatabaseClient);
+    try {
+      when(federatedDatabaseClient.get_all_databases()).thenReturn(mappedDatabases);
+    } catch (TException e) {
```
patduin: why try catch? just throw we're in a test

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -656,19 +660,61 @@ public void metastoreTunnelConfiguration() throws Exception {
...
+        .federate(SECONDARY_METASTORE_NAME, remoteServer.getThriftConnectionUri(), "remote.?database")
+        .build();
+    runWaggleDance(runner);
+
+    HiveMetaStoreClient proxy = getWaggleDanceClient();
+    List<String> allDatabases = proxy.getAllDatabases();
+
+    assertTrue(allDatabases.contains(REMOTE_DATABASE));
+
+    proxy.close();
```
patduin: why are you closing this, does it need to be. We don't do it anywhere else, probably ok.
If we need to close this we need to do it differently because this close won't be called when the test throws an exception etc...

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingServiceTest.java
AnanaMJ: I thought that if it was in a `try`, then many tests wouldn't have to be changed to include `throws MetaException, TException` but it seems that I there weren't many to change anyway.

patduin: Generally in a test, if something is wrong just throw and fail as it is probably a bug.  Perfectly fine to have `throws Exception` on every method. We want to fail as fast as possible. With you're try catch and printing test might fail and we would have to scroll to the logs to figure out why would be issue to miss.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -155,6 +169,21 @@ private void validateFederatedMetastoreDatabases(List<String> mappableDatabases,
     }
   }
 
+  private List<String> getDatabasesFromPattern(List<String> allDatabases, List<String> mappedDatabases) {
+    List<String> matchedDatabases = new ArrayList<String>();
+    for (String mappedDatabase : mappedDatabases) {
+      try {
+        Iterable<String> matches = Iterables.filter(allDatabases, Predicates.contains(Pattern.compile(mappedDatabase)));
```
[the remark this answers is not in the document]
AnanaMJ: You're right, using `Whitelist` makes it better. 
I thought that the only thing that does stuff with regex was `GrammarUtils` and didn't see how it could be used in `getDatabasesFromPattern`.

AnanaMJ:
No, there's no README change needed for this one. We specify the following:

federated-meta-stores[n].mapped-databases | No | List of databases to federate from this federated metastore, all other databases will be ignored. This property supports both full database names and (case-insensitive) [Java RegEx patterns](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).
-- | -- | --

But it was not the case for when you use MANUAL database resolution.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -155,6 +166,17 @@ private void validateFederatedMetastoreDatabases(List<String> mappableDatabases,
     }
   }
 
+  private List<String> applyWhitelist(List<String> allDatabases, List<String> mappedDatabases) {
+    List<String> matchedDatabases = new ArrayList<String>();
+    Whitelist whitelist = new Whitelist(mappedDatabases);
+    for (String database : allDatabases) {
```
massdosage: For another day but we could consider moving this method into the Whitelist class itself, especially if we find ourselves doing this filtering in multiple places.

PR #156, opened 2019-02-22
Add mechanism to execute metastore queries in parallel

AnanaMJ:


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +255,90 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
       @Override
       public List<String> getAllDatabases() {
         List<String> combined = new ArrayList<>();
-        for (DatabaseMapping mapping : databaseMappings()) {
-          try {
-            List<String> databases = mapping.getClient().get_all_databases();
-            for (String database : databases) {
-              if (isWhitelisted(mapping.getDatabasePrefix(), database)) {
-                combined.add(mapping.transformOutboundDatabaseName(database));
+        ExecutorService executorService = Executors.newCachedThreadPool();
```
patduin: We know exactly how many concurrent jobs we need so a fixed size thread pool is probably better.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +255,90 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
               }
+              LOG.info("Finished executing getAllDatabases for " + mapping.getDatabasePrefix());
+            } catch (TException e) {
+              LOG.warn("Can't fetch databases: {}", e.getMessage());
             }
-          } catch (TException e) {
-            LOG.warn("Can't fetch databases: {}", e.getMessage());
-          }
+          };
+          executorService.submit(runnableTask);
```
patduin: I think when you run this you'll see that the order of the database names coming back is not fixed anymore. This will be very confusing for a user. I think we need to have the runnable task return a results. Then loop over the Futures that are returned in order and collect the results. This way we process in parallel but returns results always in the same order.

AnanaMJ: Ah, I see. This also means that I don't need to change a lot of the tests anymore.

After I loop over the Futures, I still need to shutdown the ExecutorService, right?

patduin: yes that's good practise. To wait for a future you need to call get() I think there is a method that takes a timeout which we should use.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +258,110 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
-          try {
-            DatabaseMapping mapping = mappingWithPattern.getKey();
-            String patterns = mappingWithPattern.getValue();
-            List<TableMeta> tables = mapping.getClient().get_table_meta(patterns, tbl_patterns, tbl_types);
-            for (TableMeta tableMeta : tables) {
-              if (isWhitelisted(mapping.getDatabasePrefix(), tableMeta.getDbName())) {
-                combined.add(mapping.transformOutboundTableMeta(tableMeta));
+        Map<DatabaseMapping, String> databaseMappingsForPattern = databaseMappingsByDbPattern(db_patterns);
+        ExecutorService executorService = Executors.newFixedThreadPool(databaseMappingsForPattern.size());
+        List<Future<?>> futures = new ArrayList<Future<?>>();
```
max-jacobs: `Future<?>` could be removed to make this (a little) less verbose?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +258,110 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
+
+        for (Entry<DatabaseMapping, String> mappingWithPattern : databaseMappingsForPattern.entrySet()) {
+          Runnable runnableTask = () -> {
+            try {
+              DatabaseMapping mapping = mappingWithPattern.getKey();
+              String patterns = mappingWithPattern.getValue();
+              List<TableMeta> tables = mapping.getClient().get_table_meta(patterns, tbl_patterns, tbl_types);
+              for (TableMeta tableMeta : tables) {
+                if (isWhitelisted(mapping.getDatabasePrefix(), tableMeta.getDbName())) {
+                  combined.add(mapping.transformOutboundTableMeta(tableMeta));
```
max-jacobs: Could different threads adding to a shared list could lead to some issues? For example if one thread resizes the list while another is inserting. Maybe it would be better to use `Callable` and collect the output from each thread at the end, or perhaps there is a thread safe list implementation which can be used.

max-jacobs: Just saw Patrick's suggestion, apologies for the duplicate

AnanaMJ: Just changed `Runnable` to `Callable` hoping to preserve the order of elements added in the `combined` list (although it still doesn't fully work yet 🤔)

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
```diff
@@ -143,6 +144,14 @@ public void setWritableDatabaseWhiteList(List<String> writableDatabaseWhitelist)
     this.writableDatabaseWhitelist = writableDatabaseWhitelist;
   }
 
+  public long getTimeout() {
```
patduin: This needs to have a better name. Something like, `federatet-client-call-timeout-in-ms` I'm open for better suggestions, it should definitely tell us if the user should specify millis or seconds or minutes etc.. This needs to go in the README as well at some point

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +261,143 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
-            for (TableMeta tableMeta : tables) {
-              if (isWhitelisted(mapping.getDatabasePrefix(), tableMeta.getDbName())) {
-                combined.add(mapping.transformOutboundTableMeta(tableMeta));
+        Map<DatabaseMapping, String> databaseMappingsForPattern = databaseMappingsByDbPattern(db_patterns);
+        ExecutorService executorService = Executors.newFixedThreadPool(databaseMappingsForPattern.size());
+        List<Future<List<TableMeta>>> futures = new ArrayList<>();
+        String errorMessage = "Got exception fetching get_table_meta: ";
+
+        for (Entry<DatabaseMapping, String> mappingWithPattern : databaseMappingsForPattern.entrySet()) {
+          Callable<List<TableMeta>> callableTask = () -> {
```
patduin: can these become concrete classes?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +261,143 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
+          try {
+            DatabaseMapping mapping = iterator.next();
+            long timeout = mapping.getTimeout();
+            List<String> mappedDatabases = future.get(timeout, TimeUnit.MILLISECONDS);
+            combined.addAll(mappedDatabases);
+          } catch (InterruptedException | ExecutionException | TimeoutException | NullPointerException e) {
+            LOG.warn(errorMessage, e);
           }
         }
+        shutdownExecutorService(executorService);
```
patduin: I think this is better in a finally block

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +261,143 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
+        String errorMessage = "Can't fetch databases by pattern: ";
+
+        for (Entry<DatabaseMapping, String> mappingWithPattern : databaseMappingsForPattern.entrySet()) {
+          Callable<List<String>> callableTask = () -> {
+            try {
+              DatabaseMapping mapping = mappingWithPattern.getKey();
+              String pattern = mappingWithPattern.getValue();
+              List<String> databases = mapping.getClient().get_databases(pattern);
+              return getMappedWhitelistedDatabases(databases, mapping);
+            } catch (TException e) {
```
patduin: do we need to catch these? I think we can just throw it and they will be in the ExecutionException

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -252,62 +261,143 @@ private boolean isWhitelisted(String databasePrefix, String database) {
...
+        }
+
+        Iterator<DatabaseMapping> iterator = databaseMappingsForPattern.keySet().iterator();
+        for (Future<List<String>> future : futures) {
+          try {
+            DatabaseMapping mapping = iterator.next();
+            long timeout = mapping.getTimeout();
+            List<String> mappedDatabases = future.get(timeout, TimeUnit.MILLISECONDS);
+            combined.addAll(mappedDatabases);
+          } catch (InterruptedException | ExecutionException | TimeoutException | NullPointerException e) {
```
patduin: We need to split these up:

- InteruptException

 probably ok to log.

- ExecutionException,

 I think this is the exception which we would normally get, so we should do what we did before which is something like:
```
          } catch (ExecutionException e) {
            LOG.warn("Can't set UGI: {}", e..getCause().getMessage());
          }
```

- TimeoutException:
We want to log something else to indicate that the metastore was to slow to respond and results are omitted.

How can we get a NullPointerException?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -100,6 +106,8 @@ private void init(List<AbstractMetaStore> federatedMetaStores) {
 
   private void add(AbstractMetaStore metaStore) {
     MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(metaStore);
+    timeouts.add(metaStoreMapping.getTimeout());
```
patduin: this is not used I think

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/YamlFederatedMetaStoreStorageTest.java
```diff
@@ -154,20 +154,36 @@ public void saveFederationWriteFederations() throws Exception {
     storage.insert(newFederatedInstance);
     storage.saveFederation();
     List<String> lines = Files.readAllLines(f.toPath(), Charset.forName("UTF-8"));
-    assertThat(lines.size(), is(13));
+//    assertThat(lines.size(), is(13));
```
patduin: needs some cleanup :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: From your comment above, 

> returning null is not great I think we'll get an error adding that to the list

That's how I got NullPointerException 😬
If we don't return null, we can stop worrying about it :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
```diff
@@ -0,0 +1,128 @@
...
+import com.hotels.bdp.waggledance.mapping.model.DatabaseMapping;
+import com.hotels.bdp.waggledance.util.Whitelist;
+
+public class DatabaseMappingUtils {
+
+  public static final String MANUAL_RESOLUTION_TYPE = "MANUAL";
+  public static final String PREFIXED_RESOLUTION_TYPE = "PREFIXED";
+  private static final String INTERRUPTED_MESSAGE = "Interrupted ";
+  private static final String SLOW_METASTORE_MESSAGE = "Metastore was slow to respond so results are omitted";
+  private static final String NULL_POINTER_MESSAGE = "Got nullpointer";
```
massdosage: Got nullpointer -> Received null pointer

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
```diff
@@ -0,0 +1,128 @@
...
+    List<String> mappedDatabases = new ArrayList<>();
+    for (String database : databases) {
+      if (DatabaseMappingUtils.isWhitelisted(mapping.getDatabasePrefix(), database, mappedDbByPrefix)) {
+        mappedDatabases.add(mapping.transformOutboundDatabaseName(database));
+      }
+    }
+    return mappedDatabases;
+  }
+
+  public static void shutdownExecutorService(ExecutorService executorService) {
```
massdosage: This class is a bit of "grab bag" of functionality. The methods above are all about managing the whitelist. This one seems to be handling some boiler plate for closing an `ExecutorService` (do we even need this? is it called from more than one place?). Then the methods below seem to handle getting the results of async calls. I'd suggest splitting this into at least two classes - one with the whitelist functionality, the other with the future functionality.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private static List<?> getResultFromFuture(Iterator<DatabaseMapping> iterator, Future<List<?>> future)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    DatabaseMapping mapping = iterator.next();
+    long timeout = mapping.getTimeout();
+    return future.get(timeout, TimeUnit.MILLISECONDS);
+  }
+
+  public static void getDatabasesFromFuture(List<Future<List<?>>> futures, Iterator<DatabaseMapping> iterator,
```
massdosage: Is there a reason you pass `combined` ino this (and the other calls below) as opposed to returning a new List? (I suppose I'll find out as I read on ;) ) The reason I ask is that this is a side-effect of calling the method that isn't immediately obvious (and I think would be a functional programming no-no!) so if it's really necessary I'd suggest documenting it.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
```diff
@@ -0,0 +1,128 @@
...
+                                            List<String> combined, String errorMessage, Logger log) {
+    for (Future<List<?>> future : futures) {
+      try {
+        List<String> result = (List<String>) getResultFromFuture(iterator, future);
+        combined.addAll(result);
+      } catch (InterruptedException e) {
```
massdosage: Is there a reason you're not logging the full stack trace for any of these? Do you really need separate messages? If not you could catch all three using `|` and have a more generic `log.warn("Unexpected error getting results", e);`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
```diff
@@ -0,0 +1,128 @@
...
+        log.info(INTERRUPTED_MESSAGE);
```
massdosage: Why `info` and not `warn`? In the normal operation of things this should never happen.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
```diff
@@ -0,0 +1,128 @@
...
+      } catch (TimeoutException e) {
+        log.warn(SLOW_METASTORE_MESSAGE, e.getMessage());
+      }
+    }
+  }
+
+  public static void getTableMetaFromFuture(List<Future<List<?>>> futures, Iterator<DatabaseMapping> iterator,
+                                            List<TableMeta> combined, String errorMessage, Logger log) {
+    for (Future<List<?>> future : futures) {
+      try {
```
massdosage: This try/catch is repeated three times, you could move it into the `getResultFromFuture` method instead...

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -239,74 +247,95 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
...
 
-  private boolean isWhitelisted(String databasePrefix, String database) {
-    Whitelist whitelist = mappedDbByPrefix.get(databasePrefix);
-    if ((whitelist == null) || whitelist.isEmpty()) {
-      // Accept everything
-      return true;
-    }
-    return whitelist.contains(database);
-  }
+//  private boolean isWhitelisted(String databasePrefix, String database) {
```
massdosage: Unsure whether this should go into that MappingUtils class? If it's only used in this class then I'd say not....

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -239,74 +247,95 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
...
 
   @Override
   public PanopticOperationHandler getPanopticOperationHandler() {
     return new PanopticOperationHandler() {
 
+      private void sleepThread() {
+        try {
+          Thread.sleep(2000L);
+        } catch (InterruptedException e) {
+          LOG.info("Execution was cancelled", e.getMessage());
```
massdosage: cancelled -> interrupted
Might as well log `e` (this should never happen so having the whole stack could be useful)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -239,74 +247,95 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
...
-                combined.add(mapping.transformOutboundTableMeta(tableMeta));
-              }
-            }
-          } catch (TException e) {
-            LOG.warn("Got exception fetching get_table_meta: {}", e.getMessage());
-          }
+        Map<DatabaseMapping, String> databaseMappingsForPattern = databaseMappingsByDbPattern(db_patterns);
+        ExecutorService executorService = Executors.newFixedThreadPool(databaseMappingsForPattern.size());
+        List<Future<List<?>>> futures = new ArrayList<>();
+        String errorMessage = "Got exception fetching get_table_meta: {}";
```
massdosage: Why declare it here as a member var instead of just directly as an arg on line 288?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -239,74 +247,95 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
...
+        for (DatabaseMapping mapping : databaseMappings) {
+          GetAllDatabasesRequest allDatabasesRequest = new GetAllDatabasesRequest(mapping, mappedDbByPrefix);
+          futures.add(executorService.submit(allDatabasesRequest));
+        }
+
+        Iterator<DatabaseMapping> iterator = databaseMappings.iterator();
+        try {
+          DatabaseMappingUtils.getDatabasesFromFuture(futures, iterator, combined, errorMessage, LOG);
+        } finally {
+          DatabaseMappingUtils.shutdownExecutorService(executorService);
```
massdosage: OK, so we do call this multiple times, but I think all in this class? So just make it a method here instead of on the Utils class.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMappingTest.java
```diff
@@ -34,7 +34,7 @@
   @Test
   public void transformOutboundDatabaseName() {
     ASTQueryMapping queryMapping = ASTQueryMapping.INSTANCE;
-    MetaStoreMapping metaStoreMapping = new MetaStoreMappingImpl(PREFIX, "mapping", null, null, DIRECT);
+    MetaStoreMapping metaStoreMapping = new MetaStoreMappingImpl(PREFIX, "mapping", null, null, DIRECT, 800l);
```
massdosage: Why 8001? Looks like a magic number to me....

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: This happens for both PrefixBasedDatabaseMappingService and StaticDatabaseMappingService, since they both use an executor service that will have to be shutdown :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMappingTest.java
AnanaMJ: The number for timeout will change and will be different depending on the task executed (`setUgi`, `getTableMeta` or `getDatabases`). So, based on the graph that Patrick shared on Friday for WD tasks, we'll go with a higher limit and set that as a constant with a good, indicative name.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
AnanaMJ: Oh, you're right, this could just return a different list and then `combined` could get all of that input at the end!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
AnanaMJ: No reason, just because it was like that (e.g.
```java
catch (TException e) {
   LOG.warn("Got exception fetching get_table_meta: {}", e.getMessage());
}
```
from PrefixBasedDatabaseMappingService, master version)

I initially had it for all errors, but Patrick added [this comment](https://github.com/HotelsDotCom/waggle-dance/pull/156#discussion_r261530103) and it made sense to me :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingUtils.java
AnanaMJ: So I had it at one point like that, but if there is an error, `getResultFromFuture` would have to return null, which then threw a NullPointerException when doing `combined.addAll(null)`.

If I do a null check before I do `addAll`, I can avoid the NPE, but I'm not a fan of returning null from `getResultFromFuture`...

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: It's used in two other request classes, but I'll try to find a good place for this function :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: Fair point; I'll do that 👍

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
```diff
@@ -112,12 +110,13 @@ public void open() {
     boolean useFramedTransport = conf.getBoolVar(ConfVars.METASTORE_USE_THRIFT_FRAMED_TRANSPORT);
     boolean useCompactProtocol = conf.getBoolVar(ConfVars.METASTORE_USE_THRIFT_COMPACT_PROTOCOL);
     int clientSocketTimeout = (int) conf.getTimeVar(ConfVars.METASTORE_CLIENT_SOCKET_TIMEOUT, TimeUnit.MILLISECONDS);
+    int connectionTimeout = (int) TimeUnit.SECONDS.toMillis(2L);
```
patduin: Maybe we can define this in this class `CloseableThriftHiveMetastoreIfaceClientFactory` and pass in an argument in this class. In `CloseableThriftHiveMetastoreIfaceClientFactory` we can add `latency` from the metastore settings, that will make it configurable.

AnanaMJ: By default, the latency is 0, so all metastores that don't have a latency configured, would have connectionTimeout 0. So maybe connection timeout should be a different parameter.

patduin: no same trick as you did before, we define the base + latency

AnanaMJ: Aaah that makes sense!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -51,9 +51,9 @@
   private static final Logger LOG = LoggerFactory.getLogger(PanopticOperationHandler.class);
   private static final String INTERRUPTED_MESSAGE = "Execution was interrupted: ";
   private static final String SLOW_METASTORE_MESSAGE = "Metastore {} was slow to respond so results are omitted";
-  private static final long SET_UGI_TIMEOUT = 5000;
-  private static final long GET_DATABASES_TIMEOUT = 6000;
-  private static final long GET_TABLE_META_TIMEOUT = 400;
+  private static final long SET_UGI_TIMEOUT = 400;
```
patduin: Let's make these: `TimeUnit.MILLISECONDS.toMillis(....)` So we got units documented.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
```diff
@@ -149,7 +151,7 @@ public void open() {
           } else if (useFramedTransport) {
             transport = new TFramedTransport(transport);
           }
-          final TProtocol protocol;
+          TProtocol protocol;
```
massdosage: bye bye silly final!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -15,30 +15,82 @@
...
    * Implements {@link HMSHandler#get_database(String)} over multiple metastores
    *
    * @param databasePattern pattern to match
    * @return list of all databases that match the passed pattern
    */
-  List<String> getAllDatabases(String databasePattern);
+  public abstract List<String> getAllDatabases(String databasePattern);
+
+  protected List<String> getAllDatabases(
+      Map<DatabaseMapping, String> databaseMappingsForPattern,
```
massdosage: There's quite a lot of repeated code here for each method, we should try reduce this if possible.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -176,21 +176,23 @@ private boolean includeInResults(MetaStoreMapping metaStoreMapping) {
...
-    Iterator<Entry<String, DatabaseMapping>> iterator = mappingsByPrefix.entrySet().iterator();
-    while (iterator.hasNext()) {
-      Entry<String, DatabaseMapping> entry = iterator.next();
-      String metastorePrefix = entry.getKey();
-      if (Strings.isNotBlank(metastorePrefix) && databaseName.startsWith(metastorePrefix)) {
-        DatabaseMapping databaseMapping = entry.getValue();
-        LOG.debug("Database Name `{}` maps to metastore with prefix `{}`", databaseName, metastorePrefix);
-        if (includeInResults(databaseMapping, databaseName)) {
-          return databaseMapping;
+    synchronized (mappingsByPrefix) {
```
massdosage: 1. Why are we synchronizing? 2. and why on `mappingsByPrefix`?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -268,50 +271,52 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
         "Waggle Dance error no database mapping available tried to map database '" + databaseName + "'");
   }
 
+  @Override
+  public List<DatabaseMapping> getDatabaseMappings() {
+    Builder<DatabaseMapping> builder = ImmutableList.builder();
+    synchronized (mappingsByMetaStoreName) {
```
massdosage: Same as question above

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
```diff
@@ -49,24 +49,26 @@
...
-  private final HiveCompatibleThriftHiveMetastoreIfaceFactory hiveCompatibleThriftHiveMetastoreIfaceFactory;
-
   // for thrift connects
   private int retries = 5;
   private long retryDelaySeconds = 0;
 
-  public ThriftMetastoreClientManager(
+  private final int connectionTimeout;
+
+  ThriftMetastoreClientManager(
```
patduin: why did you remove public from this class, this is technically a backward incompatible change (it's probably fine) just wondering why

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -15,30 +15,82 @@
...
+
+    for (Entry<DatabaseMapping, String> mappingWithPattern : databaseMappingsForPattern.entrySet()) {
+      DatabaseMapping mapping = mappingWithPattern.getKey();
+      GetAllDatabasesByPatternRequest databasesByPatternRequest = new GetAllDatabasesByPatternRequest(
+          mapping, mappingWithPattern.getValue(), filter);
+      allRequests.add(databasesByPatternRequest);
+      maxLatency = Math.max(maxLatency, mapping.getLatency());
+    }
+
+    long totalTimeout = Math.max(1, GET_DATABASES_TIMEOUT + maxLatency);
```
patduin: perhaps a method: `long getTotalTimeout(GET_DATABASES_TIMEOUT , databaseMappingsForPattern.keySet())` and use that. So the maxLatency and Math.max(1...) stuff is not copied in all the places?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -15,30 +15,82 @@
...
+    List<String> result = getDatabasesByPatternFromFuture(executorService, allRequests, totalTimeout,
+        "Can't fetch databases by pattern: {}");
+    executorService.shutdownNow();
```
patduin: This should probably still be in a finally block, just in case.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: The map is synchroniszed, we're looping over it, javadoc states the pattern: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedMap-java.util.Map-

This synchronisation wasn't done in a couple of places which I think is a bug on our part, so we rectified that.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
```diff
@@ -45,14 +47,20 @@ public CloseableThriftHiveMetastoreIfaceClientFactory(
   public CloseableThriftHiveMetastoreIface newInstance(AbstractMetaStore metaStore) {
     String uris = MetaStoreUriNormaliser.normaliseMetaStoreUris(metaStore.getRemoteMetaStoreUris());
     String name = metaStore.getName().toLowerCase(Locale.ROOT);
+
+    // connection timeout should not be 0 or less
```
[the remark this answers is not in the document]
AnanaMJ: Ah, I'll correct that into 1. Should I also add an explanation as to why it's not 0?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/MetaStoreClientFactory.java
```diff
@@ -19,6 +19,10 @@
 
 public interface MetaStoreClientFactory {
 
-  CloseableThriftHiveMetastoreIface newInstance(HiveConf hiveConf, String name, int reconnectionRetries);
+  CloseableThriftHiveMetastoreIface newInstance(
```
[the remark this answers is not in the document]
AnanaMJ: Yeah, you're right. I'll do that 👍

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
patduin: ^ I guess it's sorted with the version number. (reading some comments above)

AnanaMJ: I was just going to comment to this; it was an IntelliJ change and yeah, it should be covered by the minor version change. However, this is not the aim of this PR, so I'm tempted to put it back to `public`...

patduin: it's ok

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
AnanaMJ: I would like to, but from `getAllDatabases` to `setUgi`, each time there is a for loop, a different type of object is created and it takes different parameters.

There is a lot of repetition from `getDatabasesFromFuture` to `getTableMetaFromFuture`, where the only difference is the object that is passed (all the Request classes). But I couldn't find a way to make the function work with generics :/

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
AnanaMJ: Done

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/MetaStoreClientFactory.java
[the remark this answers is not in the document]
AnanaMJ: Done

README.md
```diff
@@ -145,11 +145,13 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `primary-meta-store.writable-database-white-list`       | No       | White-list of databases used to verify write access used in conjunction with `primary-meta-store.access-control-type`. The list of databases should be listed without any `primary-meta-store.database-prefix`. This property supports both full database names and (case-insensitive) [Java RegEx patterns](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).|
 | `primary-meta-store.metastore-tunnel`                   | No       | See metastore tunnel configuration values below. |
+| `primary-meta-store.latency`                            | No       | Indicates the slowness of the metastore in **milliseconds** for increasing the default connection timeout. Default latency is `0` and should be changed if the metastore is particularly slow. If you get an error saying that result were omitted because the metastore was slow, consider changing the latency to a higher number.|
```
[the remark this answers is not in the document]
AnanaMJ: Done

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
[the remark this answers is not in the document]
AnanaMJ: Done

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
patduin: @massdosage: @AnanaMJ and me had a go to tackle some of the duplication. Should be a bit better now.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/CloseableThriftHiveMetastoreIfaceClientFactory.java
```diff
@@ -45,14 +47,21 @@ public CloseableThriftHiveMetastoreIfaceClientFactory(
   public CloseableThriftHiveMetastoreIface newInstance(AbstractMetaStore metaStore) {
     String uris = MetaStoreUriNormaliser.normaliseMetaStoreUris(metaStore.getRemoteMetaStoreUris());
     String name = metaStore.getName().toLowerCase(Locale.ROOT);
+
+    // Connection timeout should not be less than 1
+    // A timeout of zero is interpreted as an infinite timeout, so this is avoided
```
massdosage: Much clearer

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
massdosage: Looks better now, yes!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
massdosage: OK

PR #159, opened 2019-03-14
Added getPartitions() invocation logging, logs when method is called

patduin:


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -618,7 +618,7 @@ public Partition get_partition_by_name(String db_name, String tbl_name, String p
   }
 
   @Override
-  @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
+  @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME, prepend=true)
```
patduin: This will add a line in the log: `com.hotels.bdp.waggledance.server.invocation-log:168 - #get_partitions('bexg_prod_dm', 'omniture_hit_plus', -1): entered`
This is useful because `getPartitions` is a typical call that can cause issues and take a long time or never return. Normally we only log when a message returns and how long it took. This will make sure we know for certain what message was invoked.

AnanaMJ: Does `get_partitions_by_expr` also need a `prepend=true`?

patduin: I've only added it to the calls that return a `List<Partition>`, theory is that mostly those cause oom or slowness problems and are shot down before we have a change to log. That's the reason why that one and others don't have the prepend. (I have no idea if there is a big impact to prepending, but since it is off by default I thought to only add it to methods where it makes sense)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 ### Fixed
 * Support for regex `mapped-databases` for MANUAL database resolution. See [#147](https://github.com/HotelsDotCom/waggle-dance/issues/147).
 * Avoid NPE when no elements are provided for `mapped-databases` in the configuration. See [#131](https://github.com/HotelsDotCom/waggle-dance/issues/131).
+### Changed
+* Added invocation log messages for `getPartitions` calls.
```
abhimanyugupta07: is it added or changed ?

patduin: I change the annotation and the log line is added. I dunno! :)

PR #162, opened 2019-04-17
Manual client reconnection

AnanaMJ:
An issue was raised about the fact that Waggle Dance is unable to reconnect to metastores that go down and come back up when it uses a `MANUAL` database resolution. 

Example scenario that made it fail:
- Federated connection goes down
- Queries for that metastore get redirected to primary (Table not found) (expected)
- Federated connection back up
- Query still doesn't work and the metastore is not mapped anymore for Waggle Dance

The fix for this implements an observer pattern that regularly checks which connections are up and which are not and notifies `PopulateStatusFederationService` if the status (`AVAILABLE ` or `UNAVAILABLE `) changes.


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/security/AccessControlHandlerFactory.java
```diff
@@ -31,7 +32,7 @@
   private final FederationService federationService;
 
   @Autowired
-  public AccessControlHandlerFactory(FederationService federationService) {
+  public AccessControlHandlerFactory(@Qualifier("notifyingFederationService") FederationService federationService) {
```
patduin: There are 2 implementations of `FederationService` so I've added the Qualifiers to make sure we pick the correct one where we need it.

README.md
```diff
@@ -105,6 +105,7 @@ The table below describes all the available configuration values for Waggle Danc
 | `disconnect-connection-delay`     | No         | Idle metastore connection timeout. Default is `5` |
 | `disconnect-time-unit`            | No         | Idle metastore connection timeout units. Default is `MINUTES` |
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
+| `status-polling-delay`            | No         | Millisecons. Controls the delay that checks metastore status and updates long running connections of any status change. Default is `300 000` (every 5 minutes).
```
max-jacobs: `Millisecons ` -> `Milliseconds`

README.md
```diff
@@ -105,6 +105,7 @@ The table below describes all the available configuration values for Waggle Danc
 | `disconnect-connection-delay`     | No         | Idle metastore connection timeout. Default is `5` |
 | `disconnect-time-unit`            | No         | Idle metastore connection timeout units. Default is `MINUTES` |
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
+| `status-polling-delay`            | No         | Milliseconds. Controls the delay that checks metastore status and updates long running connections of any status change. Default is `300 000` (every 5 minutes).
```
abhimanyugupta07: 300 000 -> 300,000

AnanaMJ: Done :)

README.md
AnanaMJ: Done :)

README.md
patduin: hang on, let's do `300000` please as that is a valid entry. `300,000` is not. I don't think thousand separators are supported (I could be wrong). But since different countries use different thousand separators as well maybe not having one is the best option. I know we write English just think it simpler not having one. There is the human readable `every 5 minutes` already to clarify.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    // first time
+    service.poll();
+    verify(populateStatusFederationService, never()).update(primary, primary);
+    verify(populateStatusFederationService, never()).update(federate, federate);
+
+    // federated flipped status
+    federate.setStatus(MetaStoreStatus.UNAVAILABLE);
+    service.poll();
+
```
abhimanyugupta07: I feel like the verify() should happen after you have polled and once more after the next status change. It reads better and easier to understand the test.

Otherwise, you can remove the service.poll() on line 61 and write another test to check what happens if the status changes from A to U and back again to A within 5 minutes and polling service doesn't see any change.

README.md
abhimanyugupta07: Yeah. I am fine with `300000`.

AnanaMJ: Actually, yeah, you're right. The config wouldn't have comma.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/core/federation/service/PopulateStatusFederationService.java
```diff
@@ -13,23 +13,27 @@
...
+public class PopulateStatusFederationService implements FederationService {
 
-  private final FederationService federationService;
+  private final NotifyingFederationService federationService;
   private final FederationStatusService federationStatusService;
 
-  public DelegatingFederationService(
-      FederationService federationService,
+  public PopulateStatusFederationService(
+      NotifyingFederationService federationService,
```
max-jacobs: Would it be better to use the interface here to decouple these two classes?

```java
public PopulateStatusFederationService(
      FederationService federationService,
      FederationStatusService federationStatusService)

You may not even have to configure the bean because the only beans that can be injected here are this class itself and the `NotifyingFederationService`, so Spring might be able to inject for you.

File: waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java [hunk 6]
```diff
@@ -0,0 +1,71 @@
...
+    primary.setStatus(MetaStoreStatus.AVAILABLE);
+    federate.setStatus(MetaStoreStatus.AVAILABLE);
+
+    List<AbstractMetaStore> metastores = Lists.newArrayList(primary, federate);
+    when(populateStatusFederationService.getAll()).thenReturn(metastores);
+
+    // first time
+    service.poll();
+    verify(populateStatusFederationService, never()).update(primary, primary);
+    verify(populateStatusFederationService, never()).update(federate, federate);
```
max-jacobs: On the first run, the `previous` map should only contain null values, and therefore the metastore statuses should not be checked. To verify this, `primary` and `federate` could be mocks. Then, instead of modifying their statuses, the behaviour of `primaryMock.getStatus()` and `federateMock.getStatus()` could be changed. Just a thought :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
max-jacobs: I agree, i find its easier to read a long test like this if the expected behaviour is explicitly verified as you go through. What do others think?

AnanaMJ: So would it be easier to read if we have the following tests:
* one metastore changes the status once and we verify that update was called after `service.poll()`
* one primary and one federated metastore change status at the same time (one from `UNAVAILABLE` to `AVAILABLE` and the other one, the other way around) and we verify that update is called for both after `service.poll()`
* one metastore changes status like this `AVAILABLE` &rightarrow; `UNAVAILABLE` &rightarrow; `AVAILABLE` with no `service.poll()` and we verify that there were no interactions

?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
AnanaMJ: Hmm we are checking the status [on this line](https://github.com/HotelsDotCom/waggle-dance/pull/162/files#diff-8998b00790a07c6ea9c5d45816f75e6fR48). So we should check that it's checked once instead of twice, right?

max-jacobs: You're right. And actually we don't really need to test this, this wouldn't be testing behaviour, but implementation of that behaviour. For example, the use of `getStatus()` could be extracted to a variable and then it will be used once anyway. The check that no update is made is definitely enough, so forget what I said, apologies.

AnanaMJ: Do you mean something like this?
```java
when(populateStatusFederationService.getAll()).thenReturn(Collections.singletonList(federateMock));
when(federateMock.getStatus()).thenReturn(MetaStoreStatus.AVAILABLE);

// first time
service.poll();
verify(populateStatusFederationService, never()).update(federateMock, federateMock);
verify(federateMock).getStatus();

//second time
when(federateMock.getStatus()).thenReturn(MetaStoreStatus.UNAVAILABLE);
service.poll();
verify(populateStatusFederationService).update(federateMock, federateMock);
verify(federateMock, times(3)).getStatus(); // 1 the first time and 2 the second time
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/core/federation/service/PopulateStatusFederationService.java
AnanaMJ: Will confirm if this works and Spring doesn't add `PopulateStatusFederationService` instead of `NotifyingFederationService` after I add some tests 😆

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
max-jacobs: I don't think it is necessary to split this into different tests. Would be easier to read if we verify what updates are done directly after the polls, including verifying that no updates are done to primary on the second poll, and not verifying twice at the end.

If you want to split it up, I would say keep the test exactly as it is, essentially:
- One poll, verify no updates are made
- Poll, don't change statuses, poll again and verify that no updates are made
- Poll, do change statuses, poll again and verify that updates are made

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/core/federation/service/PopulateStatusFederationService.java
AnanaMJ: Actually, IntelliJ gives an autowire error when I make the change from `NotifyingFederationService` to `FederationService`:

>      Could not autowire. There is more than one bean of 'FederationService' type.
>      Beans: notifyingFederationService    (NotifyingFederationService.java)
>             populateStatusFederationService    (PopulateStatusFederationService.java)

patduin: you need to add @Qualifier("notifyingFederationService") if you want to use the interface

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
AnanaMJ: Done :) I kept them in the same test.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/core/federation/service/PopulateStatusFederationService.java
AnanaMJ: Thanks, that worked :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
patduin: @max-jacobs I had what you described but changed it to do the last verify at the end: `verify(populateStatusFederationService, times(2)).update(federate, federate);` 
Because that last verify verifies everything so it as the times(2).
Which then looks a bit weird because it seems that the last poll does the update twice. 
With the amount of comments in the test I definitely think it should be rewritten. It's obviously not clear to everyone :)

README.md
```diff
@@ -105,6 +105,7 @@ The table below describes all the available configuration values for Waggle Danc
 | `disconnect-connection-delay`     | No         | Idle metastore connection timeout. Default is `5` |
 | `disconnect-time-unit`            | No         | Idle metastore connection timeout units. Default is `MINUTES` |
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
+| `status-polling-delay`            | No         | Milliseconds. Controls the delay that checks metastore status and updates long running connections of any status change. Default is `300000` (every 5 minutes).
```
massdosage: metastore status -> metastore availability [or connectivity, I'm fine with either]

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfiguration.java
```diff
@@ -45,6 +45,7 @@
   private int thriftServerStopTimeoutValInSeconds = 60;
   private int thriftServerRequestTimeout = 20;
   private TimeUnit thriftServerRequestTimeoutUnit = TimeUnit.SECONDS;
+  private Long statusPollingDelay = null;
```
massdosage: Why is this modelled as a `Long` and not a `long` with a default like the other similar values above?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationService.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final PopulateStatusFederationService populateStatusFederationService;
+  private Map<String, MetaStoreStatus> previous = new HashMap<>();
+
+  public PollingFederationService(PopulateStatusFederationService populateStatusFederationService) {
+    this.populateStatusFederationService = populateStatusFederationService;
+  }
+
+  // status-polling-delay, default every 5 minutes
+  @Scheduled(fixedDelayString = "${status-polling-delay:300000}")
```
massdosage: OK, so there is a default here, not sure what the pros/cons are of having it here as opposed to in the 'WaggleDanceConfiguration'

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfiguration.java
```diff
@@ -118,4 +119,12 @@ public void setThriftServerRequestTimeoutUnit(TimeUnit thriftServerRequestTimeou
...
   }
 
+  public Long getStatusPollingDelay() {
+    return statusPollingDelay;
+  }
+
+  public void setStatusPollingDelay(Long statusPollingDelay) {
+    this.statusPollingDelay = statusPollingDelay;
+  }
+
```
massdosage: To be consistent with `disconnect-connection-delay` shouldn't we also have a TimeUnit for this rather than forcing ms?

waggle-dance/pom.xml
```diff
@@ -25,6 +26,17 @@
       <version>${project.version}</version>
       <classifier>exec</classifier>
     </dependency>
+    <dependency>
+      <groupId>com.hotels</groupId>
+      <artifactId>waggle-dance-rest</artifactId>
+      <version>${project.version}</version>
+      <exclusions>
```
massdosage: Why do we need the exclusion? Isn't everything here at the same version?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfiguration.java
patduin: default is in the annotation

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationService.java
patduin: we'd have the default value in two places. If I remove it here you'd get a weird error if you set it to null. Doesn't matter much, I found this easiest

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfiguration.java
patduin: Then we can't use the annotation anymore. The spring schedule doesn't have a TimeUnit

waggle-dance/pom.xml
patduin: classifier is different

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfiguration.java
massdosage: Bugger, that's a pity as we now have an inconsistent way of configuring "time related things". I guess the alternative would be that we do it the consistent way here and then convert it into ms before we call spring? If we only allow it to be configured here and then set it later that would be nice but I see your point on the annotation, you're saying it can't be given a value other than via a property? Would we have to programmatically do some stuff in the constructor or somewhere else instead?

waggle-dance/pom.xml
massdosage: Hmm, OK, so even though we specifically depend on the `exec` classifier of core above this is then pulling in another version of it transitively?

README.md
AnanaMJ: Done

waggle-dance/pom.xml
patduin: yeah it's a different artifact as far as maven is concerned.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/conf/WaggleDanceConfiguration.java
patduin: We could not sue the annotations and make a schedule bean/class/thing programatically (I think). Then we can control it better. I felt that was overkill but I don't mind doing it.

massdosage: Yeah, I think since a lot of our "contract" is actually what end users see in the yaml file we should make that as simple and consistent as possible. As an end user I think I'd find it confusing if some times are specified with a unit and some not, I'd prefer either one or the other wouldn't you? If you agree then I think we should do the extra work in the code to make this better for them.

README.md
```diff
@@ -105,6 +105,7 @@ The table below describes all the available configuration values for Waggle Danc
 | `disconnect-connection-delay`     | No         | Idle metastore connection timeout. Default is `5` |
 | `disconnect-time-unit`            | No         | Idle metastore connection timeout units. Default is `MINUTES` |
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
+| `status-polling-delay`            | No         | Milliseconds. Controls the delay that checks metastore availability and updates long running connections of any status change. Default is `300000` (every 5 minutes).
```
massdosage: This needs updating to reflect the changes just committed

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/context/ScheduledBeansTest.java
```diff
@@ -0,0 +1,63 @@
...
+    final AtomicInteger pollCallCount = new AtomicInteger(0);
+    doAnswer(new Answer<Void>() {
+      @Override
+      public Void answer(InvocationOnMock invocation) throws Throwable {
+        pollCallCount.incrementAndGet();
+        log.info("test poll called");
+        return null;
+      }
+    }).when(pollingFederationService).poll();
+    if (pollCallCount.get() == 0) {
```
massdosage: Why not use awaitility here instead? await.atMost(1 second).until(bla) <- something like that?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/context/ScheduledBeansTestContext.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ScheduledBeansTestContext {
+
+  @Bean
+  public WaggleDanceConfiguration waggleDanceConfiguration() {
+    WaggleDanceConfiguration mock = Mockito.mock(WaggleDanceConfiguration.class);
+    when(mock.getStatusPollingDelay()).thenReturn(10);
+    when(mock.getStatusPollingDelayTimeUnit()).thenReturn(TimeUnit.MILLISECONDS);
+    return mock;
+
```
massdosage: The nefarious blank line!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/context/ScheduledBeans.java
```diff
@@ -0,0 +1,57 @@
...
+      WaggleDanceConfiguration waggleDanceConfiguration,
+      PollingFederationService pollingFederationService) {
+    this.waggleDanceConfiguration = waggleDanceConfiguration;
+    this.pollingFederationService = pollingFederationService;
+  }
+
+  @Override
+  public void configureTasks(ScheduledTaskRegistrar taskRegistrar) {
+    Runnable task = new Runnable() {
+      @Override
```
max-jacobs: Waggle Dance is java 8 so I think we could use a lambda here if its preferred? I don't mind if not

patduin: good catch! lamdafied it!

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/context/ScheduledBeansTest.java
patduin: done

README.md
```diff
@@ -105,6 +105,8 @@ The table below describes all the available configuration values for Waggle Danc
 | `disconnect-connection-delay`     | No         | Idle metastore connection timeout. Default is `5` |
 | `disconnect-time-unit`            | No         | Idle metastore connection timeout units. Default is `MINUTES` |
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
+| `status-polling-delay`            | No         | Controls the delay that checks metastore availability and updates long running connections of any status change. Default is `5` (every 5 minutes).
```
AnanaMJ: I think there should be a `|` at the end of this line and the next one.

PR #169, opened 2019-04-30
adding prefixes for UDF's used in view

### LINKED ISSUE #165: Prefix UDF's referenced in a view correctly.
_As a user of Waggle Dance
I'd like the udfs to be correctly prefixed if they are defined in a view_

Doing a `describe extented my_view` on a table that is a view will show the create statement for that view. Databases are correctly prefixed when of the form `<database>.<table>` but for udfs they are not. The `<database>.<udf>` is not prefixed. This means that querying the view will result in an error like: `FAILED: SemanticException [Error 10011]: Invalid function <database>.<udf>` because the database is not prefixed.

patduin:
fixes #165

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
@@ -23,6 +23,8 @@ import java.util.List;
 import java.util.SortedSet;
 import java.util.Stack;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
```diff
@@ -37,6 +39,7 @@
 
   INSTANCE;
 
+  private static final String RE_WORD_BOUNDARY = "\\b";
```
max-jacobs: Here we have `static final String ` whereas below we have `final static Comparator`, so a slight difference, I think `static final` is the convention so shall we change the `Comparator`?

PR #170, opened 2019-05-02
Fix rest api

patduin:
Added a wd-boot module to capture the dependency on core and rest, the only purpose of the module is to repackage itself as a spring boot service executable jar.

### DIFF
--- a/waggle-dance-boot/pom.xml
+++ b/waggle-dance-boot/pom.xml
@@ -10,7 +10,6 @@
 
   <artifactId>waggle-dance-boot</artifactId>
   <description>Waggle Dance Spring boot module</description>
-  <packaging>jar</packaging>
 
   <properties>
     <spring-boot-maven-plugin.version>1.5.7.RELEASE</spring-boot-maven-plugin.version>
@@ -46,10 +45,6 @@
               <groupId>org.apache.hive</groupId>
               <artifactId>hive-common</artifactId>
             </dependency>
-<!--            <dependency>
-              <groupId>com.hotels</groupId>
-              <artifactId>waggle-dance-core</artifactId>
-            </dependency>-->
           </requiresUnpack>
           <executable>true</executable>
           <classifier>exec</classifier>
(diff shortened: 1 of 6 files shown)


waggle-dance-boot/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+              setting fixes that problem.
+            -->
+            <dependency>
+              <groupId>org.apache.hive</groupId>
+              <artifactId>hive-common</artifactId>
+            </dependency>
+<!--            <dependency>
+              <groupId>com.hotels</groupId>
+              <artifactId>waggle-dance-core</artifactId>
+            </dependency>-->
```
AnanaMJ: I guess this can be removed?

waggle-dance/src/main/assembly/waggle-dance-package.xml
```diff
@@ -16,10 +16,10 @@
   <dependencySets>
     <dependencySet>
       <outputDirectory>service</outputDirectory>
-      <outputFileNameMapping>${artifact.artifactId}-latest${dashClassifier?}.${artifact.extension}</outputFileNameMapping>
+      <outputFileNameMapping>waggle-dance-core-latest${dashClassifier?}.${artifact.extension}</outputFileNameMapping>
```
massdosage: Anyway we could achieve this with symlinks for backwards compatability with the old "core" names but pointing to files with names not containing core (the "boot" files)?

patduin: not really. can't do symlinks in assembly

massdosage: Maven Ant plugin here we come!

patduin: the name will be `waggle-dance-core-latest-exec.jar` not the same as the normal "core" jar it will have the exec classifier. I agree not ideal to have "core" in the name but I rather not clutter it all up with symlinks and just keep it backward compatible.

massdosage: OK.

PR #171, opened 2019-05-09
Added panoptic handler of get_all_functions

### LINKED ISSUE #164: Show functions from all metastores
_As a user of Waggle Dance
I'd like to see all functions (not just the primary metastore functions)
So I can more easily use access UDFs across metastores._

Acceptance Criteria:
* `show functions` will show all functions from all metastores (correctly prefixing UDFs).

patduin:
fixes #164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## TBD
 ### Fixed
-* Show Functions now shows UDF's from all metastores. See [#164](https://github.com/HotelsDotCom/waggle-dance/issues/164).
+* `Show Functions` now shows UDFs from all metastores. See [#164](https://github.com/HotelsDotCom/waggle-dance/issues/164).
 * Fixed REST API (http://localhost:18000/api/admin/federations/) which broke in 3.3.0 release.
-* Prefixing of UDF's used in a view. See [#165](https://github.com/HotelsDotCom/waggle-dance/issues/165).
+* Prefixing of UDFs used in a view. See [#165](https://github.com/HotelsDotCom/waggle-dance/issues/165).
 
 ## [3.3.0] - 2019-04-30
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## TBD
 ### Fixed
+* Show Functions now shows UDF's from all metastores. See [#164](https://github.com/HotelsDotCom/waggle-dance/issues/164).
 * Fixed REST API (http://localhost:18000/api/admin/federations/) which broke in 3.3.0 release.
 * Prefixing of UDF's used in a view. See [#165](https://github.com/HotelsDotCom/waggle-dance/issues/165).
 
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticConcurrentOperationExecutor.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticConcurrentOperationExecutor.java
@@ -1,0 +1,95 @@
+/**
+ * Copyright (C) 2016-2019 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.mapping.service;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.hotels.bdp.waggledance.mapping.model.DatabaseMapping;
+import com.hotels.bdp.waggledance.mapping.service.requests.RequestCallable;
+
+public class PanopticConcurrentOperationExecutor implements PanopticOperationExecutor {
+
+  private static final String INTERRUPTED_MESSAGE = "Execution was interrupted: ";
+  private static final String SLOW_METASTORE_MESSAGE = "Metastore {} was slow to respond so results are omitted";
+
+  private final static Logger LOG = LoggerFactory.getLogger(PanopticConcurrentOperationExecutor.class);
+
+  @Override
+  public <T> List<T> executeRequests(
+      List<? extends RequestCallable<List<T>>> allRequests,
+      long requestTimeout,
+      String errorMessage) {
+    ExecutorService executorService = Executors.newFixedThreadPool(allRequests.size());
+    try {
+      List<T> allResults = new ArrayList<>();
+      List<Future<List<T>>> futures = Collections.emptyList();
+      Iterator<? extends RequestCallable<List<T>>> iterator = allRequests.iterator();
+
+      try {
+        long totalTimeout = getTotalTimeout(requestTimeout, allRequests);
+        futures = executorService.invokeAll(allRequests, totalTimeout, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
+        LOG.warn("Execution was interrupted", e);
+      }
+
+      for (Future<List<T>> future : futures) {
+        DatabaseMapping mapping = iterator.next().getMapping();
+        List<T> result = getResultFromFuture(future, mapping.getMetastoreMappingName(), errorMessage);
+        allResults.addAll(result);
+      }
+      return allResults;
+    } finally {
+      executorService.shutdownNow();
+    }
+  }
+
+  private <T> List<T> getResultFromFuture(Future<List<T>> future, String metastoreMappingName, String errorMessage) {
+    try {
+      return future.get();
+    } catch (InterruptedException e) {
+      LOG.warn(INTERRUPTED_MESSAGE, e);
+    } catch (ExecutionException e) {
+      LOG.warn(errorMessage, e.getCause().getMessage());
+    } catch (CancellationException e) {
+      LOG.warn(SLOW_METASTORE_MESSAGE, metastoreMappingName);
+    }
+    return Collections.emptyList();
+  }
+
+  private <T> long getTotalTimeout(long requestTimeout, List<? extends RequestCallable<List<T>>> allRequests) {
+    long maxLatency = Integer.MIN_VALUE;
+    for (RequestCallable<List<T>> request : allRequests) {
+      maxLatency = Math.max(maxLatency, request.getMapping().getLatency());
+    }
+
+    // Connection timeout should not be less than 1
+    // Other implementations interpret a timeout of zero as infinite wait
+    // `Future.get` currently does not do that, but this is safe if implementation changes in the future
+    return Math.max(1, requestTimeout + maxLatency);
+  }
+}
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
@@ -198,9 +198,7 @@ public abstract class PanopticOperationHandler {
    *
    * @return GetAllFunctionsResponse (db's from federated metastores will be prefixed if necessary)
    */
-  public abstract GetAllFunctionsResponse getAllFunctions();
-
-  protected GetAllFunctionsResponse getAllFunctions(List<DatabaseMapping> databaseMappings) {
+  public GetAllFunctionsResponse getAllFunctions(List<DatabaseMapping> databaseMappings) {
     List<GetAllFunctionsRequest> allRequests = new ArrayList<>();
 
     for (DatabaseMapping mapping : databaseMappings) {
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
@@ -16,32 +16,22 @@
 package com.hotels.bdp.waggledance.mapping.service;
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.concurrent.CancellationException;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 
 import org.apache.hadoop.hive.metastore.HiveMetaStore.HMSHandler;
 import org.apache.hadoop.hive.metastore.api.GetAllFunctionsResponse;
 import org.apache.hadoop.hive.metastore.api.TableMeta;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.hotels.bdp.waggledance.mapping.model.DatabaseMapping;
 import com.hotels.bdp.waggledance.mapping.service.requests.GetAllDatabasesByPatternRequest;
 import com.hotels.bdp.waggledance.mapping.service.requests.GetAllFunctionsRequest;
 import com.hotels.bdp.waggledance.mapping.service.requests.GetTableMetaRequest;
-import com.hotels.bdp.waggledance.mapping.service.requests.RequestCallable;
 import com.hotels.bdp.waggledance.mapping.service.requests.SetUgiRequest;
 
 /**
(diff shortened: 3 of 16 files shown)


waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
```diff
@@ -445,6 +446,41 @@ public void panopticOperationsHandlerSetUgiDifferentTimeouts() throws Exception
...
+
+    PanopticOperationHandler handler = service.getPanopticOperationHandler();
+    GetAllFunctionsResponse result = handler.getAllFunctions();
+    assertThat(result.getFunctionsSize(), is(2));
+    assertThat(result.getFunctions().get(0).getFunctionName(), is("fn1"));
+    assertThat(result.getFunctions().get(1).getFunctionName(), is("fn2"));
+  }
+
+  @Test
+  public void panopticOperationsHandlerGetAllFunctionsWithSlowConnection() throws Exception {
```
patduin: There is a lot of duplication in these tests, I'll have a look and see if I can refactor it a bit.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## TBD
 ### Fixed
+* Show Functions now shows UDF's from all metastores. See [#164](https://github.com/HotelsDotCom/waggle-dance/issues/164).
```
AnanaMJ: -> "`show functions` now shows UDFs" 
(no apostrophe for plurals unless it's for single letters or single numbers - I only know this because I googled it 😄)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -188,4 +192,38 @@
     // `Future.get` currently does not do that, but this is safe if implementation changes in the future
     return Math.max(1, requestTimeout + maxLatency);
   }
+
+  /**
+   * Implements {@link HMSHandler#get_all_functions()} over multiple metastores
+   *
+   * @return GetAllFunctionsResponse (db's from federated metastores will be prefixed if necessary)
+   */
+  public abstract GetAllFunctionsResponse getAllFunctions();
```
AnanaMJ: I think this function and the one below could be moved on line 143, to be in the same bundle with the other functions that implement an `HMSHandler` function.
Not fussed about it, but it might be useful in the future when someone looks at this class :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -188,4 +192,38 @@
...
+      allRequests.add(getAllFunctionsRequest);
+    }
+
+    List<GetAllFunctionsResponse> responses = executeRequests(allRequests, GET_ALL_FUNCTIONS_TIMEOUT,
+        "Got exception fetching get_all_functions: {}");
+    if (responses.isEmpty()) {
+      return new GetAllFunctionsResponse();
+    }
+    GetAllFunctionsResponse result = new GetAllFunctionsResponse(responses.get(0));
+    responses.stream().skip(1).forEach(response -> addFunctionsToResult(response, result));
```
max-jacobs: The `1` confused me, but then I realised the first response is being added through the constructor in the line before. Could also use the default constructor and flatmap it, I _think_ this works:

```responses.stream().flatMap(response -> response.getFunctions().stream()).forEach(function -> result.addToFunctions(function));```

Then wouldn't need to method below. I think null functions can also be filtered too.

File: waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java [hunk 5]
```diff
@@ -333,9 +334,15 @@ public PanopticOperationHandler getPanopticOperationHandler() {
...
+
+      @Override
+      public GetAllFunctionsResponse getAllFunctions() {
+        List<DatabaseMapping> databaseMappings = getDatabaseMappings();
+        return super.getAllFunctions(databaseMappings);
+      }
     }
 
-        ;
+    ;
```
max-jacobs: Could this go on line 343 like in `PrefixBasedDatabaseMappingService`?

AnanaMJ: Should this be near the curly brace at line 343?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -333,9 +334,15 @@ public PanopticOperationHandler getPanopticOperationHandler() {
         }
         return combined;
       }
+
+      @Override
+      public GetAllFunctionsResponse getAllFunctions() {
+        List<DatabaseMapping> databaseMappings = getDatabaseMappings();
+        return super.getAllFunctions(databaseMappings);
```
AnanaMJ: Since this is the same as the one in `PrefixedDatabaseMappingService`, do you think it's worth adding these lines in `PanopticOperationHandler`'s `getAllFunctions()` abstract class and remove the `abstract`?

I'm asking this because I know we did the same for `setUgi`. However, if it's better to have the implementations in the concrete classes and leave just the abstracts in `PanopticOperationHandler`, then it might be good to move the implementation of `setUgi` in `PrefixedDatabaseMappingService` and here too.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
AnanaMJ: Wouldn't that add the first response twice though? Because `result` already has the first response.

max-jacobs: Yes sorry, this is only with the default constructor. Like so:

```java
GetAllFunctionsResponse result = new GetAllFunctionsResponse();
responses.stream()
    .flatMap(response -> response.getFunctions()
        .stream().filter(functions -> functions != null))
    .forEach(function -> result.addToFunctions(function));
```

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -202,6 +209,36 @@ public void typical() throws Exception {
     // Remote table
     String waggledRemoteDbName = REMOTE_DATABASE;
     assertTypicalRemoteTable(proxy, waggledRemoteDbName);
+
```
AnanaMJ: blank line 🙂

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -202,6 +209,36 @@ public void typical() throws Exception {
...
+    localServer.client().createFunction(localFunction);
+    Function remoteFunction = new Function("fn2", REMOTE_DATABASE, "com.hotels.hive.FN1", "hadoop", PrincipalType.USER,
+        0, FunctionType.JAVA, resourceUris);
+    remoteServer.client().createFunction(remoteFunction);
+
+    GetAllFunctionsResponse allFunctions = proxy.getAllFunctions();
+    List<Function> functions = allFunctions.getFunctions();
+    assertThat(functions.size(), is(2));
+    assertThat(functions.get(0).getFunctionName(), is("fn1"));
+    assertThat(functions.get(0).getDbName(), is(LOCAL_DATABASE));
```
AnanaMJ: If primary also has a prefix, then this should return the prefixed version of the local database, right?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
AnanaMJ: I see, thanks!
Although, if the default constructor is used, I think it's easier to understand what `addFunctionsToResult` does without looking at the implementation instead of looking at 
`flatMap(response -> response.getFunctions().stream().filter(functions -> functions != null))`

I'm not saying the `flatMap` is hard to understand, just that the function describes what it does pretty well.

max-jacobs: Even without the `flatmap` I don't think the `.skip(1)` is necessary, can still just use the default constructor with no `skip`. But yes, true, the method name makes it easier to convey the meaning. Happy with either way :)

patduin: I want to use the copy constructor just in case anything else besides the Functions are copied there. I want it to be like: Create a new response like the response from the primary (first one) then add all functions to this new response.

I'll have a closer look at Max suggestion not sure if I find it easier to read.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
patduin: yeah I'll move it a bit I think I'm gonna change this class a bit to make it more testable

CHANGELOG.md
patduin: yeah I thought so but it looked weirder :) I'll change it :)

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
patduin: yes

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
patduin: yes good point

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
patduin: @max-jacobs  your proposal filter out nulls from the functions list but the actual list can be null.

This I think would be same as my original code. Using `response.isSetFunctions()` but that's the same as `response.getFunctions != null`
```
    responses
        .stream()
        .skip(1)
        .filter(response -> response.isSetFunctions())
        .flatMap(response -> response.getFunctions().stream())
        .forEach(function -> result.addToFunctions(function));
```

max-jacobs: That looks good. If you don't think its easier to read then by all means stick with whats there currently. Happy with either way, I just love a good stream :)

patduin: I'm going for it!

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/PanopticConcurrentOperationExecutorTest.java
```diff
@@ -0,0 +1,138 @@
...
+    assertThat(executeRequests.get(1), is("call2"));
+  }
+
+  @Test
+  public void executeRequestsExceptionLoggedResultsReturned() throws Exception {
+    PanopticConcurrentOperationExecutor executor = new PanopticConcurrentOperationExecutor();
+    DummyRequestCallable errorRequest = new DummyRequestCallable("call2", mapping2) {
+      @Override
+      public List<String> call() throws Exception {
+        throw new RuntimeException("who you gonna call...");
```
AnanaMJ: Ghostbusters! 🚫 👻

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/PanopticConcurrentOperationExecutorTest.java
```diff
@@ -0,0 +1,138 @@
...
+      };
+    };
+    List<DummyRequestCallable> allRequests = Lists
+        .newArrayList(new DummyRequestCallable("call1", mapping1), errorRequest);
```
AnanaMJ: What if we make `call1` return an error instead of `call2`? This way we can check that 1 failed request won't affect all of them.

patduin: calls are done concurrently right? how is that different? I'll add a call3 to cover all bases, makes it more clear

AnanaMJ: Oh, actually you're right. They are all resolved at the same time with `invokeAll`. This should be fine then 😄

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -297,9 +299,15 @@ public PanopticOperationHandler getPanopticOperationHandler() {
...
         }
-        List<String> result = executeRequests(allRequests, GET_DATABASES_TIMEOUT, "Can't fetch databases: {}");
+        List<String> result = getPanopticOperationExecutor()
+            .executeRequests(allRequests, GET_DATABASES_TIMEOUT, "Can't fetch databases: {}");
         return result;
       }
+
+      @Override
+      protected PanopticOperationExecutor getPanopticOperationExecutor() {
+        return new PanopticConcurrentOperationExecutor();
```
massdosage: Is it OK/necessary to create a new one every time?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/requests/GetAllFunctionsRequest.java
```diff
@@ -0,0 +1,50 @@
...
+  private final DatabaseMapping mapping;
+
+  public GetAllFunctionsRequest(DatabaseMapping mapping) {
+    this.mapping = mapping;
+  }
+
+  @Override
+  public List<GetAllFunctionsResponse> call() throws Exception {
+    GetAllFunctionsResponse response = mapping.getClient().get_all_functions();
+    if (response.getFunctions() != null) {
```
massdosage: I think you need a test for this returning null, and maybe the next line returning empty?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/PanopticConcurrentOperationExecutorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    when(mapping1.getMetastoreMappingName()).thenReturn("mapping1");
+    when(mapping2.getMetastoreMappingName()).thenReturn("mapping2");
+    when(mapping3.getMetastoreMappingName()).thenReturn("mapping3");
+  }
+
+  @Test
+  public void executeRequestsInOrder() throws Exception {
+    PanopticConcurrentOperationExecutor executor = new PanopticConcurrentOperationExecutor();
+    List<DummyRequestCallable> allRequests = Lists
+        .newArrayList(new DummyRequestCallable("call1", mapping1), new DummyRequestCallable("call2", mapping2));
```
massdosage: string sort order could theoretically be why they are sorted below so introduce a call that doesn't sort.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/stubs/HiveStubs.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.hadoop.hive.metastore.api.ResourceUri;
+
+import com.google.common.collect.Lists;
+
+public final class HiveStubs {
+
+  private HiveStubs() {}
+
+  public static Function newFunction(String databaseName, String functionName) {
+    ArrayList<ResourceUri> resourceUris = Lists
```
massdosage: Concrete type on LHS OMG!

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -118,6 +125,7 @@ public void init() throws Exception {
 
   @After
   public void destroy() throws Exception {
+    graphite.after();
```
massdosage: Maybe for a separate PR?

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -204,6 +212,35 @@ public void typical() throws Exception {
...
+    runner = WaggleDanceRunner
+        .builder(configLocation)
+        .databaseResolution(DatabaseResolution.PREFIXED)
+        .primary("primary", localServer.getThriftConnectionUri(), READ_ONLY)
+        .federate(SECONDARY_METASTORE_NAME, remoteServer.getThriftConnectionUri(), REMOTE_DATABASE)
+        .build();
+
+    runWaggleDance(runner);
+    HiveMetaStoreClient proxy = getWaggleDanceClient();
+    ArrayList<ResourceUri> resourceUris = Lists
```
massdosage: Again....

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: it is OK it's just the object creation the actual method called is the same as before so just object creation overhead which is fine probably even preferred for every request etc. Keeps things simple

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
patduin: yes was just running a quick test. (didn't work)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticConcurrentOperationExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.hotels.bdp.waggledance.mapping.model.DatabaseMapping;
+import com.hotels.bdp.waggledance.mapping.service.requests.RequestCallable;
+
+public class PanopticConcurrentOperationExecutor implements PanopticOperationExecutor {
+
+  private static final String INTERRUPTED_MESSAGE = "Execution was interrupted: ";
+  private static final String SLOW_METASTORE_MESSAGE = "Metastore {} was slow to respond so results are omitted";
+
+  private final static Logger LOG = LoggerFactory.getLogger(PanopticConcurrentOperationExecutor.class);
```
massdosage: LOG -> log please (since it's not immutable)

patduin: We use LOG everywhere in WD, don't mind changing but I think better to stick with the same

massdosage: Ah, OK, I lose track of which projects we do which in. Agree, fine for another PR to make it consistent everywhere, I can just picture @AnanaMJ loving that one!

PR #174, opened 2019-05-23
removed check for prefix so we can fallback to empty

### LINKED ISSUE #173: Fallback to primary metastore when no prefix
In #152 we added the feature where the primary metastore can have a prefix. However this change will not be backwards compatible if we enable this for an existing Apiary installation where users are used to having their queries succeed against the primary without having a prefix. There might be workarounds we can do with configuring the metastore multiple times with and without a prefix but this will lead to duplicate tables etc. being returned.

Ideally we want this situation to prevail:
* standard name is with prefix in all cases - i.e. all calls which return values have the prefixed version and the non-prefixed version does not need to be returned.
* backward compatible feature is this: “if query does not have prefix then WD defaults to using the prefix associated with the primary metastore”.  That does require all current WD implementations to do a 1 time update of their configs when deploying this feature (as prefix in config for primary metastore is required).  However, one change, one notice to users, automatic backward compat, and things are standardized from then on. 

Given that at some point in the future we want to remove the concept of the primary metastore  (see #154) perhaps it would be worth introducing a flag that you can put on the primary metastore like `fallback-metastore=true` or `fallback-prefix=""` so that it can be put on *any* metastore but only *one* per WD instance.

patduin:
fixes #173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 ### Changed
 * Changed a prefixed *primary* metastore to fallback to 'empty prefix' if nothing specified. See [#173](https://github.com/HotelsDotCom/waggle-dance/issues/173).
 
-
 ## [3.3.1] - 2019-05-20
 ### Fixed
 * `Show Functions` now shows UDFs from all metastores. See [#164](https://github.com/HotelsDotCom/waggle-dance/issues/164).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [3.3.2] - TBD
+### Changed
+* Changed a prefixed *primary* metastore to fallback to 'empty prefix' if nothing specified. See [#173](https://github.com/HotelsDotCom/waggle-dance/issues/173).
+
+
 ## [3.3.1] - 2019-05-20
 ### Fixed
 * `Show Functions` now shows UDFs from all metastores. See [#164](https://github.com/HotelsDotCom/waggle-dance/issues/164).
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingImpl.java
@@ -78,11 +78,10 @@ class MetaStoreMappingImpl implements MetaStoreMapping {
   @Override
   public String transformInboundDatabaseName(String databaseName) {
     databaseName = databaseName.toLowerCase(Locale.ROOT);
-    if (!databaseName.startsWith(getDatabasePrefix())) {
-      throw new IllegalArgumentException(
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## [3.3.2] - TBD
+### Changed
+* Changed a prefixed *primary* metastore to fallback to 'empty prefix' if nothing specified. See [#173](https://github.com/HotelsDotCom/waggle-dance/issues/173).
+
+
```
AnanaMJ: Two new lines 😱

AnanaMJ: (although it's not visible in the formatted version :) )

patduin:
...to easy indeed. Found a conundrum in UDFs. 
Leaving out the prefix for a udf won't work. At least not in a Hive CLI (Not sure how this works in other clients (Spark,Qubole etc..)). Hive CLI fetches all UDFs on startup (or `reload function` command). They are registered *with* prefix. So when calling `show functions` you get `prefix_db.functionName`.
In order for a `select db.functionName` to work we'll need to return that particular function name. We need to return both `prefix_db.functionName` and `db.functionName` when loading the functions. This means show functions will show both which is something we don't really want.
I don't think there is any way around this though when using a function (e.g. `select...`) there is no Metastore call to proxy it just looks up the loaded functions in the client and fails if it cannot locate a function with that name.

We need to choose either list functions prefixed and non-prefixed *or* force prefixed udfs only. Thoughts/preferences anyone?

abhimanyugupta07:
I think I prefer showing both in `show functions` rather than having inconsistent behaviour when compared with rest of the queries where we are defaulting to `db` when no prefix is provided in the query.

patduin:
> I think I prefer showing both in `show functions` rather than having inconsistent behaviour when compared with rest of the queries where we are defaulting to `db` when no prefix is provided in the query.

The problem with showing both (same when showing both db's) is that it looks weird and user might think hey that looks wrong I'll delete on (which will delete both). :(

abhimanyugupta07:
> > I think I prefer showing both in `show functions` rather than having inconsistent behaviour when compared with rest of the queries where we are defaulting to `db` when no prefix is provided in the query.
> 
> The problem with showing both (same when showing both db's) is that it looks weird and user might think hey that looks wrong I'll delete on (which will delete both). :(

True. But deleting a function is not as big of a problem compared to deleting a table or database. Out of the two options, if you choose the one where you force people to use prefix, again the issue of backward compatibility will arise and also a user will be confused that for everything else my query works but for functions, its not working. I feel it's inconsistent from an end-user perspective. Also (I am guessing here) `show functions` might not be a commonly used query, so this might not be that big of a problem if we show both options only for this query.

massdosage:
Remind me what happens when one does "show databases" in this case - do we only see the prefixed ones?

patduin:
> Remind me what happens when one does "show databases" in this case - do we only see the prefixed ones?

yes.

PR #176, opened 2019-07-23
Show mapped databases for primary

### LINKED ISSUE #175: Filter DB list in primary-meta-store
https://groups.google.com/forum/#!topic/waggle-dance-user/mpwQ9U6Lr2I

Is it possible to filter the DB list in primary-meta-store? I tried the below, but still i get to see the tables not mapped in primary.
primary-meta-store:
  access-control-type: READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST
  database-prefix: ''
  latency: 0
  name: primary
  remote-meta-store-uris: thrift://127.0.0.1:9083
  mapped-databases:
  - userdatalake
  writable-database-white-list:
  - userdatalake
federated-meta-stores:
- access-control-type: READ_ONLY
  database-prefix: waggle_prod_
  latency: 0
  mapped-databases:
  - test
  - users
  name: waggle_prod
  remote-meta-store-uris: thrift://10.25.10.10:9083

$ hive --hiveconf hive.metastore.uris=thrift://localhost:9083 -e "show databases;"
userdatalake
proddatalake
qedatalake

$ hive --hiveconf hive.metastore.uris=thrift://10.25.10.10:9083 -e "show databases;"
test
users
devdatalake
mldatalake

$ hive --hiveconf hive.metastore.uris=thrift://localhost:48869 -e "show databases;"
userdatalake
proddatalake
qedatalake
waggle_prod_test
waggle_prod_users

Is it possible for the above command to show only the filtered DB as below?
$ hive --hiveconf hive.metastore.uris=thrift://localhost:48869 -e "show databases;"
userdatalake
waggle_prod_test
waggle_prod_users

Only userdatalake from Primary should be visible.

Thanks,
Anil

AnanaMJ:
Fixes #175 

---

Extra tests done (for prefixed and manual mapping):

**Config 1**:
```yaml
primary-meta-store:
  access-control-type: READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST
  database-prefix: ''
  name: primary
  remote-meta-store-uris: thrift://metastore-uri:port
  mapped-databases:
  - test
  writable-database-white-list:
  - test
```

* `show databases` returns only `test`
* if I try to do `show tables in other_test;` (where `other_test` exists, but is not mapped), I get `FAILED: SemanticException [Error 10072]: Database does not exist: other_test`
* `show tables in test;` returns all tables in `test`

**Config 2** (no mapped databases):
```yaml
primary-meta-store:
  access-control-type: READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST
  database-prefix: ''
  name: primary
  remote-meta-store-uris: thrift://metastore-uri:port
  writable-database-white-list:
  - test
```

* `show databases` returns all databases from metastore

**Config 3** (empty mapped databases):
```yaml
primary-meta-store:
  access-control-type: READ_AND_WRITE_AND_CREATE_ON_DATABASE_WHITELIST
  database-prefix: ''
  name: primary
  remote-meta-store-uris: thrift://metastore-uri:port
  mapped-databases: []
  writable-database-white-list:
  - test
```

* `show databases` returns nothing
* attempting to query a database, the same `database does not exist` exception comes up


waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStoreTest.java
```diff
@@ -119,4 +119,21 @@ public void constructorWithArrayListForWhtelist() {
...
     assertThat(store.getWritableDatabaseWhiteList(), is(whitelist));
   }
+
+
+  @Test
+  public void mappedDatabases() {
+    List<String> mappedDatabases = new ArrayList<>();
+    mappedDatabases.add("database");
+    metaStore.setMappedDatabases(mappedDatabases);
+    assertThat(metaStore.getMappedDatabases(), is(mappedDatabases));
```
abhimanyugupta07: Add an assert for the boolean too: `metastore.shouldHaveNoMappedDatabases()`

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStoreTest.java
```diff
@@ -119,4 +119,21 @@ public void constructorWithArrayListForWhtelist() {
...
+  }
+
+  @Test
```
abhimanyugupta07: Add a test for empty too coz you are handling the empty list a bit differently in your setter.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingService.java
```diff
@@ -33,7 +35,7 @@
    * @param databaseName given database name
    * @return the {@link DatabaseMapping} that maps to the given databaseName
    */
-  DatabaseMapping databaseMapping(@NotNull String databaseName);
+  DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSuchObjectException;
```
abhimanyugupta07: Just for my information, why this change was required ?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -80,32 +81,40 @@ public PrefixBasedDatabaseMappingService(
     init(initialMetastores);
   }
 
-  private void init(List<AbstractMetaStore> federatedMetaStores) {
+  private void init(List<AbstractMetaStore> abstractMetaStores) {
```
abhimanyugupta07: I don't think you need to rename this variable but if you do, name it `metastores` and subsequent usage should be `metastore` instead of `abstractMetastore`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -80,32 +81,40 @@ public PrefixBasedDatabaseMappingService(
...
+    mappedDbByPrefix.put(metaStoreMapping.getDatabasePrefix(), mappedDbWhitelist);
   }
 
-  private Whitelist getWhitelistedDatabases(FederatedMetaStore federatedMetaStore) {
-    return new Whitelist(federatedMetaStore.getMappedDatabases());
+  private Whitelist getWhitelistedDatabases(AbstractMetaStore metaStore) {
+    List<String> databasesPatternToMap;
+
+    // check if metastore is supposed to have all databases matched or not
+    if (!metaStore.shouldHaveNoMappedDatabases() && metaStore.getMappedDatabases().isEmpty()) {
```
abhimanyugupta07: Do you need to check both these things here ? I think `shouldHaveNoMappedDatabases()` is supposed to handle both null and empty usecase.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -171,9 +172,21 @@ private void validateFederatedMetastoreDatabases(List<String> mappableDatabases,
     }
   }
 
-  private List<String> applyWhitelist(List<String> allDatabases, List<String> mappedDatabases) {
+  private Whitelist getWhitelistedDatabases(AbstractMetaStore metaStore) {
+    List<String> databasesPatternToMap;
+
+    // check if metastore is supposed to have all databases matched or not
+    if (!metaStore.shouldHaveNoMappedDatabases() && metaStore.getMappedDatabases().isEmpty()) {
```
abhimanyugupta07: Same question here.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -323,17 +338,20 @@ public PanopticOperationHandler getPanopticOperationHandler() {
...
           }
-          combined.addAll(mappingsByDatabaseName.keySet());
-        } catch (ExecutionException e) {
-          LOG.warn("Can't fetch databases: {}", e.getCause().getMessage());
         }
-        return combined;
+
+        Collections.sort(primaryCombined);
+        Collections.sort(federatedCombined);
+        primaryCombined.addAll(federatedCombined);
```
abhimanyugupta07: why not sort the final list once before returning ? is this intentional ?

AnanaMJ: If I sort it at the end, then I would get the order of databases like ['federated_db', 'primary_db'] and that's not an expected behaviour in the tests right now. I wasn't sure about this one, but if someone does manual mapped databases, would they be ok with combining the federated databases with the primary databases?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: You're right, the first check should be enough. Thanks :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -257,7 +274,7 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) {
 
   private boolean isWhitelisted(String databasePrefix, String database) {
     Whitelist whitelist = mappedDbByPrefix.get(databasePrefix);
-    if ((whitelist == null) || whitelist.isEmpty()) {
```
patduin: was this a bug before?

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStore.java
```diff
@@ -16,13 +16,16 @@
...
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import javax.validation.constraints.NotNull;
 
 public class PrimaryMetaStore extends AbstractMetaStore {
 
   private static final String EMPTY_PREFIX = "";
+  private @NotNull List<String> mappedDatabases = Collections.emptyList();
```
patduin: I still think it can be null as it is an optional thing. The default should be null which is accept all. It shouldn't be emptyList which means map nothing.

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStore.java
```diff
@@ -47,6 +50,24 @@ public FederationType getFederationType() {
...
   }
 
+  @Override
+  public List<String> getMappedDatabases() {
+    return mappedDatabases;
+  }
+
+  @Override
+  public void setMappedDatabases(List<String> mappedDatabases) {
+    if (mappedDatabases != null) {
```
patduin: Not sure if this is needed,

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
```diff
@@ -152,6 +152,12 @@ public void setLatency(long latency) {
     this.latency = latency;
   }
 
+  public abstract List<String> getMappedDatabases();
```
patduin: isn't the logic in the subclasses the same, can we move it here instead of abstract?

AnanaMJ: If we move it there, then when the yaml file is overwritten, if the original yaml file had no line for mapped databases, then it will be overwritten to `mapped-databases: []`

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStore.java
AnanaMJ: If I don't have it, I get a nullpointer exception instead of a validation error for `mapped-databases: `

AnanaMJ: Because it does `null.isEmpty()`

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -80,32 +81,40 @@ public PrefixBasedDatabaseMappingService(
...
-  private void add(AbstractMetaStore federatedMetaStore) {
-    MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(federatedMetaStore);
+  private void add(AbstractMetaStore metaStore) {
+    MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(metaStore);
 
-    if (federatedMetaStore.getFederationType() == PRIMARY) {
+    if (metaStore.getFederationType() == PRIMARY) {
       primaryDatabaseMapping = createDatabaseMapping(metaStoreMapping);
       mappingsByPrefix.put(metaStoreMapping.getDatabasePrefix(), primaryDatabaseMapping);
+      primaryShouldMatchAllDatabases =
```
patduin: we don't need this boolean for the federated metastore why can't we handle the primary case the same way. I think we need to handle this in the Whitelist better, e.g. allow getMappedDatabases to return null. The whitelist should be created with null and it should accept everything.

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStore.java
patduin: Why don't we treat `mapped-databases:` as null
This weird stuff with the if something is not set we treat it as null via this boolean but we actually return an emptyList is just very confusing. :)

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
patduin: not if we allow null values, which I think we should

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStore.java
patduin: With not needed I mean the whole if block :). Just set the value and leave all logic out of this class. I feel it belongs in the Whitelist class that should deal with null, empty etc...

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
```diff
@@ -152,6 +152,12 @@ public void setLatency(long latency) {
...
+
+  public abstract void setMappedDatabases(List<String> mappedDatabases);
+
+  public abstract boolean shouldHaveNoMappedDatabases();
```
patduin: I think this doesn't end up in the yml, could you check if it ends up in the json? http://localhost:18000/api/admin/federations/

AnanaMJ: It doesn't (which is good, right?), but `"mappedDatabases":[]` does

patduin: yes

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
AnanaMJ: Hmm I'm not sure, but if whitelist is empty now, we don't want to accept anything. Now, if whitelist is empty, it means that we need to reject everything.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
abhimanyugupta07: @patduin may be able to confirm if there is a requirement to display databases  in this particular manner or they can all be in a sorted manner ?

patduin: We should tamper with the order, whatever order we got back from the metastore should be preserved. Then we usually do, primary first, followed by federated metastores (as configured in the yml).

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/security/ReadWriteCreateAccessControlHandler.java
```diff
@@ -29,7 +46,27 @@ public boolean hasCreatePermission() {
...
+    // Notify to update mapped databases
+    List<String> mappedDatabases = null;
+    if (metaStore.getMappedDatabases() != null) {
+      mappedDatabases = new ArrayList<>(metaStore.getMappedDatabases());
+      if (!mappedDatabases.contains(name)) {
+        mappedDatabases.add(name);
+      }
+    }
+
+    AbstractMetaStore newMetaStore;
```
patduin: All this code should be inside the `if (metaStore.getMappedDatabases() != null) {` that will avoid unnecessary updates.

AnanaMJ: Even if it's null, we still need to do `federationService.update(metaStore, newMetaStore);` because we want `mappingsByDatabaseName` to contain the new database.

patduin: ah yes you are right

AnanaMJ: I only found out because a test was failing 😄  _(thank you, team standards XD)_

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -323,17 +317,7 @@ public PanopticOperationHandler getPanopticOperationHandler() {
...
-          List<String> databases = primaryDatabasesCache.get(PRIMARY_KEY);
-          for (String database : databases) {
-            combined.add(primaryDatabaseMapping.transformOutboundDatabaseName(database));
-          }
-          combined.addAll(mappingsByDatabaseName.keySet());
-        } catch (ExecutionException e) {
-          LOG.warn("Can't fetch databases: {}", e.getCause().getMessage());
-        }
-        return combined;
+        return new ArrayList<>(mappingsByDatabaseName.keySet());
```
patduin: I think there is some risk now that the you won't see a database added to the primary database (if it was added outside of WD). If a client reconnects it will see the up-to-date list. So probably fine.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -94,54 +95,53 @@ public StaticDatabaseMappingService(
...
-      }
       validateFederatedMetastoreDatabases(mappableDatabases, metaStoreMapping);
-      DatabaseMapping databaseMapping = createDatabaseMapping(metaStoreMapping);
-      mappingsByMetaStoreName.put(metaStoreMapping.getMetastoreMappingName(), databaseMapping);
-      addDatabaseMappings(mappableDatabases, databaseMapping);
     }
+
+    mappingsByMetaStoreName.put(metaStoreMapping.getMetastoreMappingName(), databaseMapping);
+    addDatabaseMappings(mappableDatabases, databaseMapping);
+    databaseMappingToDatabaseList.put(databaseMapping, mappableDatabases);
```
patduin: I think we should do this put on the key: `databaseMapping.getMetastoreMappingName()`. So we don't rely on the object instance equals of a DataMapping object.

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -784,4 +824,63 @@ public void getDatabaseFromPatternPrefixed() throws Exception {
     List<String> expected = Lists.newArrayList("default", LOCAL_DATABASE, PREFIXED_REMOTE_DATABASE);
     assertThat(allDatabases, is(expected));
   }
+
+  @Test
+  public void primaryMappedDatabasesManual() throws Exception {
+    localServer.createDatabase("random_primary");
```
patduin: <3 nice

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
AnanaMJ: Good point, thanks :)

waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/PrimaryMetaStoreTest.java
```diff
@@ -119,4 +119,5 @@ public void constructorWithArrayListForWhtelist() {
     assertThat(store.getAccessControlType(), is(accessControlType));
     assertThat(store.getWritableDatabaseWhiteList(), is(whitelist));
   }
+
```
abhimanyugupta07: There is a minor spelling mistake in the function name: `constructorWithArrayListForWhtelist` -> `constructorWithArrayListForWhitelist`

AnanaMJ: Done :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -80,32 +80,26 @@ public PrefixBasedDatabaseMappingService(
...
     }
   }
 
-  private void add(AbstractMetaStore federatedMetaStore) {
-    MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(federatedMetaStore);
+  private void add(AbstractMetaStore metaStore) {
+    MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(metaStore);
 
-    if (federatedMetaStore.getFederationType() == PRIMARY) {
+    if (metaStore.getFederationType() == PRIMARY) {
```
abhimanyugupta07: I am not exactly sure how `primaryDatabaseMapping` is used in rest of the WD code but perhaps this can be re-written like this:

```
 DatabaseMapping databaseMapping = createDatabaseMapping(metaStoreMapping);

    }
    mappingsByPrefix.put(metaStoreMapping.getDatabasePrefix(), databaseMapping);
```

AnanaMJ: You're right, that would be much simpler; thanks :)
It's done now

PR #180, opened 2019-07-30
refactor code

chergey:
Hi,
I decided replace LoadingCache in StaticDatabaseMappingService with memoizing supplier which better conveys its purpose and more suited for the task, I think. Besides, removed throws clauses.

What do you think ?


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/ExpiringMemoizingSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Adapted from
+ *
+ * @see com.google.common.base.Suppliers.ExpiringMemoizingSupplier
+ * to support invalidation
+ */
+
+public interface ExpiringMemoizingSupplier<T> extends Supplier<T>, Serializable {
```
patduin: Not really happy that we have to have this class in here I think I prefer the cache in that case, why we can't use the ExpriringMemoizingSupplier from Guava?
Really like the java8 facelift otherwise :)

chergey: I think, cache should be used for caching, not for memoization. Guava memoizing supplier doesn't support invalidation.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -194,27 +194,27 @@ public void shutdown() {
 
   @Override
   @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
-  public String getMetaConf(String key) throws MetaException, TException {
+  public String getMetaConf(String key) throws TException {
```
massdosage: OK, so the exception signatures on this class come from implementing `org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore` which has the `MetaException` even thought it extends `TException`. Same for the methods below. We could remove them but I think I'd actually prefer to keep them as they're more self-documenting which I assume is one of the reasons they were kept in the original interface. I'd rather we stick with what the class we're implementing does, even if it's not perfect.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -164,18 +164,18 @@
...
-  private DatabaseMapping checkWritePermissions(String databaseName) throws TException {
+  private DatabaseMapping checkWritePermissions(String databaseName) {
     DatabaseMapping mapping = databaseMappingService.databaseMapping(databaseName);
     mapping.checkWritePermissions(databaseName);
     return mapping;
   }
 
   @Override
-  public void close() throws IOException {
+  public void close() {
```
massdosage: This is overriding the method in `java.io.Closeable` which does throw an `IOException` so we should leave this.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -1667,7 +1646,7 @@ public GetTableResult get_table_req(GetTableRequest req) throws MetaException, N
   @Override
   @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
   public GetTablesResult get_table_objects_by_name_req(GetTablesRequest req)
-      throws MetaException, InvalidOperationException, UnknownDBException, TException {
+      throws TException {
```
massdosage: I'm sorry but I think we should revert the changes to this entire class so it's more in line with the class its implementing.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/ExpiringMemoizingSupplier.java
massdosage: I'd also prefer to use the cache rather than introducing a new class with a whole bunch of hand-crafted functionality (that isn't tested) that we need to maintain. Even if a cache isn't the best fit at least its common library code in Guava that has been battle-hardened and use in production all over the show.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -15,100 +15,78 @@
...
+                        } catch (TException e) {
+                            throw new WaggleDanceException("Can't validate database clashes", e.getCause());
+                        }
+                    } else {
+                        return Lists.newArrayList();
+                    }
+                }, 1, TimeUnit.MINUTES);
+
+
+        mappingsByMetaStoreName = Collections.synchronizedMap(new LinkedHashMap<>());
```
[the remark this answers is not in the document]
chergey: It was done to make fields final.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
chergey: Ok

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/ExpiringMemoizingSupplier.java
chergey: How about I write a test for it ?

massdosage: The problem is we then have to maintain this forever so I'd rather not. If this was fixing a bug or adding some new feature it would be different story but purely for an "idiomatic" change I don't think it's worth the burden on the maintainers of the project in future.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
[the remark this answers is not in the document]
massdosage: OK, it's only a few lines so it doesn't make a big difference in this case.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/ExpiringMemoizingSupplier.java
chergey: I see your reason however I don't see any trouble in maintaining a trivial utility class. If is of such importance for you, I shall remove all the commits and leave only those that you find acceptable.

massdosage: Yes please, thanks!

massdosage:
Could you also please resolve the merge conflicts?

chergey:
I removed unnecessary changes and resolved conflicts.

chergey:
IDE - Intellij IDEA. You are using something else ?

massdosage:
> 
> 
> IDE - Intellij IDEA. Are you using something else ?

We use a mix of Eclipse and Intellij. @AnanaMJ explained she ran some automated cleanup tasks too.

massdosage:
Not sure why Travis is being an idiot and failing some of the tests. I ran the whole build locally and it succeeded. Will merge now and see what our internal Jenkins CI/CD pipeline thinks of it. Thanks for the contribution!

PR #182, opened 2019-09-05
Added logging and clarified README

patduin:


README.md
```diff
@@ -425,6 +425,8 @@ Being a Spring Boot Application, all [standard actuator endpoints](https://docs.
 
 eg. Healthcheck Endpoint: http://localhost:18000/actuator/health
 
+On top of the Spring endpoints there are endpoints available to get more details information. All the urls are printed when Waggle Dance starts up. Most notable is: `http://host:18000//api/admin/federations`. Which returns information about the availability of the metastores that are used. It can be used for troubleshooting it is not recommended to use that as a healthcheck.
```
AnanaMJ: "more details information" -> "more detailed information"
"urls" -> "URLs"
Last sentence would work with a "but" in the middle: "It can be used for troubleshooting, but it is not recommended ..."

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TSetIpAddressProcessorFactory.java
```diff
@@ -47,6 +53,9 @@ public TSetIpAddressProcessorFactory(
   @Override
   public TProcessor getProcessor(TTransport transport) {
     try {
+      Socket socket = ((TSocket) transport).getSocket();
```
massdosage: I *think* the way we have this all set up this will always be of type `TSocket` so this should never throw a `ClassCastException` right? Alternative would be to do an `instanceof` check first...

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TSetIpAddressProcessorFactory.java
```diff
@@ -47,6 +53,9 @@ public TSetIpAddressProcessorFactory(
...
+      log.debug("Opened a connection from ip: {}", socket.getInetAddress().getHostAddress());
```
massdosage: Opened a connection from ip -> Received a connection from IP
(I think that reads a bit better, NABD either way)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TSetIpAddressProcessorFactory.java
patduin: yes, I didn't bother with the check. Should be alright.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TSetIpAddressProcessorFactory.java
patduin: I agree

PR #184, opened 2019-09-26
Allow federated metastores to have empty prefix

### LINKED ISSUE #183: Allowing federated metastores to take on an empty string for prefix
**Is your feature request related to a problem? Please describe.**
I am trying to have a pre-prod setup for my spark scripts - they are supposed to read from prod storage and write to pre-prod storage. Hence, I want to have the federated metastore point to my prod, and primary metastore pointing to my pre-prod. To minimize code change required in the spark scripts, I want to call the databases as they are in prod, but the configuration does not allow me to leave the federated metastore prefix empty.

**Describe the solution you'd like**
I'd like to be able to leave the federated metastore prefix as an empty string for at least one metastore. Perhaps this could be done by just doing a check on all prefixes and approving them as long as they are unique, rather than hardcoding a standard.

**Describe alternatives you've considered**
No working alternatives to this issue.

AnanaMJ:
Fixes #183


AnanaMJ:
@patduin Added a line :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/YamlFederatedMetaStoreStorageTest.java
```diff
@@ -89,6 +89,20 @@ public void loadFederation_singleFederation() throws Exception {
...
+    storage.loadFederation();
+  }
+
+  @Test (expected = IllegalArgumentException.class)
+  public void loadTwoEmptyPrefixes() throws Exception {
+    File f = dataFolder.getFile("two-empty-prefixes.yml");
+    YamlFederatedMetaStoreStorage storage = new YamlFederatedMetaStoreStorage(f.toURI().toString(), configuration);
+    storage.loadFederation();
+  }
+
```
massdosage: Is it worth having a test of the "happy day" scenario where have a federated metastore with no prefix and a primary with a prefix? Just to validate that this is allowed and the resultant metastores have the correct prefixes?

AnanaMJ: Yeah, I think it's worth testing that it actually works in a yaml file. I've just added it :)

PR #191, opened 2020-03-02
Prometheus

max-jacobs:
### :link: Related Issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ### Changed
 * Updated `hotels-oss-parent` to 5.0.0 (was 4.0.1).
 
+### Added
+* Support for prometheus metrics.
+
 ## [3.5.0] - 2019-10-14
 ### Added
 * Added logging to help debug connection issues.
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ The following snippet shows a typical Graphite configuration:
 | `graphite.poll-time`                                  | No       | Amount of time between Graphite polls. Defaults to `5000`. |
 | `graphite.poll-time-unit`                             | No       | Time unit of `graphite.poll-time` - this is [the list of allowed values](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/TimeUnit.html). Defaults to `MILLISECONDS`. |
 
-Prometheus can also be used to gather metrics. This can be done by enabling the Prometheus endpoint in the configuration:
+Prometheus can also be used to gather metrics. This can be done by enabling the Prometheus endpoint `/prometheus` in the configuration:
 
     management.endpoints.web.exposure.include: prometheus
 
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 ### Changed
 * Updated `hotels-oss-parent` to 5.0.0 (was 4.0.1).
 
+### Added
+* Support for prometheus metrics.
```
massdosage:
```suggestion
* Support for Prometheus metrics.
```

waggle-dance-core/pom.xml
```diff
@@ -97,6 +97,11 @@
       <artifactId>micrometer-registry-graphite</artifactId>
     </dependency>
 
+    <dependency>
```
massdosage: To be really pedantic this should go below the jmx one if we want to keep these alpha-sorted by groupid.artifactid where possible.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/metrics/MonitoringConfigurationTestContext.java
```diff
@@ -27,4 +28,10 @@ GraphiteConfiguration graphiteConfiguration() {
     return conf;
   }
 
+  @Bean
+  PrometheusConfiguration prometheusConfiguration() {
```
massdosage: So, just to confirm, we now *always* expose a Prometheus endpoint? i.e. it's not something you can turn on or off?

max-jacobs: No I don't think so. The new Prometheus dependency will automatically create the PrometheusRegistry bean only. To expose the endpoint, when the jar is started up with some configuration, that configuration will need to include which endpoints to expose - this addition will happen here I think https://github.com/ExpediaGroup/apiary-federation/tree/master/templates

...subject to testing of course

massdosage: OK, make sense, we should document this then as per my other comment.

README.md
```diff
@@ -310,6 +310,10 @@ The following snippet shows a typical Graphite configuration:
 | `graphite.poll-time`                                  | No       | Amount of time between Graphite polls. Defaults to `5000`. |
 | `graphite.poll-time-unit`                             | No       | Time unit of `graphite.poll-time` - this is [the list of allowed values](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/TimeUnit.html). Defaults to `MILLISECONDS`. |
 
+Prometheus can also be used to gather metrics. This can be done by enabling the Prometheus endpoint in the configuration:
```
patduin: can you add there what the default "url" would be will just make it a little easier to find I assume it's normally printed startup logs?

max-jacobs: I think the default url is just `/prometheus`

max-jacobs: Done

patduin: ah ok that trivial, excellent :)

Issue #195, opened 2020-08-10
Metastores should support "mapped-tables" similar to "mapped-databases"


barnharts4:
**Is your feature request related to a problem? Please describe.**
When federating certain metastores, I need a way to only expose certain tables from certain schemas in the metastore.

**Describe the solution you'd like**
I would like a configuration parameter like `mapped-tables` or `mapped-db-tables` that would work similar to `mapped-databases` (or supersede `mapped-databases`) that would allow Regexs to be used to specify a combination of mapped schemas and tables from the schemas.

**Describe alternatives you've considered**
 Currently we are managing this with Apache Ranger policies to only allow access to certain tables in the schemas we are federating.  However, the downside to this is that they are still visible, leading customer to try to query them, and Ranger policies can be somewhat complex to create if they span multiple schemas/tables/users.

andreeapad:
Released in Waggle Dance 3.8.0

PR #196, opened 2020-08-20
Feature/database name mapping

patduin:


README.md
```diff
@@ -149,13 +149,15 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `primary-meta-store.latency`                            | No       | Indicates the acceptable slowness of the metastore in **milliseconds** for increasing the default connection timeout. Default latency is `0` and should be changed if the metastore is particularly slow. If you get an error saying that results were omitted because the metastore was slow, consider changing the latency to a higher number.|
 | `primary-meta-store.mapped-databases`                   | No       | List of databases to federate from the primary metastore; all other databases will be ignored. This property supports both full database names and [Java RegEx patterns](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) (both being case-insensitive). By default, all databases from the metastore are federated. |
+| `federated-meta-stores[n].database-name-mapping`             | No       | BiDirectional Map of database names and mapped names. key=<database name as known in the primary metastore>, value=<name that should be shown to a client>. See the [Database Name Mapping](#database-name-mapping) section.|
```
JayGreeeen, line 152: Looks like this is in here twice, and the info between `< >` isnt showing up in the file.

README.md
```diff
@@ -424,7 +426,45 @@ Assumes database resolution is done by adding prefixes. If database resolution i
      where h.date = '2016-05-13'
        and h.hour = 1
     ;
-    
+
+## Database Name Mapping
+Waggle Dance allows configuration to map database names. This feature is intended to be useful when migrating from an existing databases into different environments.
```
JayGreeeen, line 431:
```suggestion
Waggle Dance allows configuration to map database names. This feature is intended to be useful when migrating from an existing database into different environments.
```

README.md
```diff
@@ -424,7 +426,45 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+A use case:
+We have legacy data lakes that have tables belonging to the same domain in differently names databases. E.G a booking domain is called 'datawarehouse' or 'hotel_booking'. When federating to those datalakes it helps data discovery if those names are standardized to one name. Renaming a database is easy but migrate ETL might not be so to easy migration you can configure Waggle Dance to remap the names.
```
JayGreeeen, line 433:
```suggestion
We have legacy data lakes that have tables belonging to the same domain in differently named databases. E.g. a booking domain is called 'datawarehouse' or 'hotel_booking'. When federating to those datalakes it helps data discovery if those names are standardized to one name. Renaming a database is easy but migrating ETL might not be, so to make migration easier you can configure Waggle Dance to remap the names.
```

README.md
```diff
@@ -424,7 +426,45 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+Example Datalake X federating to 2 other "legacy" datalakes Y and Z. 
+This is how we want the different databases to look like in the datalakes:
```
JayGreeeen, line 436:
```suggestion
This is what we want the different databases to look like in the datalakes:
```

README.md
```diff
@@ -424,7 +426,45 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+        - datawarehouse
+        database-name-mapping:
+          datawarehouse: booking
+      - remote-meta-store-uris: thrift://10.0.0.2:9083
+        name: z
+        database-name-mapping:
+          hotel_booking: booking
+ 
+   
+If an optional `mapped-databases` is used that filter is applied first the renaming is applied after.
```
JayGreeeen, line 465:
```suggestion
If an optional `mapped-databases` is used that filter is applied first and the renaming is applied after.
```

README.md
```diff
@@ -424,7 +426,45 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+NOTE: mapping names adds an extra layer of abstraction and we advice to use this as temporary migration solutions only. It becomes harder to debug where a virtual (remapped) table actually is coming from.
```
JayGreeeen, line 466:
```suggestion
NOTE: mapping names adds an extra layer of abstraction and we advise to use this as a temporary migration solution only. It becomes harder to debug where a virtual (remapped) table actually is coming from.
```

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -880,4 +882,68 @@ public void primaryMappedDatabasesPrefixed() throws Exception {
...
+        .withPrimaryDatabaseNameMappingMap(databaseNameMapping1)
+        .federate(SECONDARY_METASTORE_NAME, remoteServer.getThriftConnectionUri(), REMOTE_DATABASE)
+        .withFederatedDatabaseNameMappingMap(databaseNameMapping2)
+        .build();
+
+    runWaggleDance(runner);
+    HiveMetaStoreClient proxy = getWaggleDanceClient();
+
+    List<String> allDatabases = proxy.getAllDatabases();
+    System.out.println(allDatabases);
```
JayGreeeen, line 904: :o

waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -880,4 +882,68 @@ public void primaryMappedDatabasesPrefixed() throws Exception {
...
```
JayGreeeen, line 936: :oo

README.md
patduin: first one should be primary, I'll have a look

massdosage:
Is this intended to fix #195?

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
      where h.date = '2016-05-13'
        and h.hour = 1
     ;
+
+## Database Name Mapping
+Waggle Dance allows configuration to map database names. This feature is intended to be useful when migrating from an existing database into different environments.
+A use case:
+We have legacy data lakes that have tables belonging to the same domain in differently named databases. E.g. a booking domain is called 'datawarehouse' or 'hotel_booking'. When federating to those datalakes it helps data discovery if those names are standardized to one name. Renaming a database is easy but migrating ETL might not be, so to make migration easier you can configure Waggle Dance to remap the names.
+
+Example Datalake X federating to 2 other "legacy" datalakes Y and Z.
```
massdosage, line 435:
```suggestion
So in this example we have a new Data lake X which is federated along with 2 other "legacy" data lakes Y and Z.
```

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+This is what we want the different databases to look like in the datalakes:
```
massdosage, line 436:
```suggestion
The desired end result is to be able to refer to all the booking related database as "booking" prefixed by the source data lake name.
```

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+X has:
```
massdosage, line 438: X doesn't actually have these right, it's what we want the names to look like from Waggle Dance? X might have its own booking database I guess? Not sure if we should put that in the example, might make it more confusing?

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+    y_booking
+    z_booking
+
+Y has:
+    
+    datawarehouse
```
massdosage, line 445:
```suggestion
    datawarehouse (this actually contains "booking" tables but is misnamed according to the new convention)
```

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+    another_db
```
massdosage, line 446:
```suggestion
    another_db (this contains tables we want to refer to using this existing name, i.e. we don't want to change this)
```

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+Z has:
     
+    hotel_booking
```
massdosage, line 450:
```suggestion
    hotel_booking (this also contains "booking" tables but is misnamed according to the new convention)
```

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+To achieve this we can configure Waggle Dance in X to map the names:
```
massdosage, line 452:
```suggestion
To achieve a unified view of all the booking tables in the different databases (without actually renaming them in Hive) we can configure Waggle Dance in X to map from the old to the new names like so:
```

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+        mapped-databases:
+        - datawarehouse
+        database-name-mapping:
+          datawarehouse: booking
+      - remote-meta-store-uris: thrift://10.0.0.2:9083
+        name: z
+        database-name-mapping:
+          hotel_booking: booking
+
+
```
massdosage, line 467: This now lets you refer to the above referenced databases in scripts as either "y_datawarehouse" or "y_booking" and "z_hotel_booking" or "z_booking".

massdosage: I'm correct in that you can use both the old and new names right?

Not sure whether it would be useful to show an example bit of SQL demonstrating before and after to make this all a bit clearer?

README.md
```diff
@@ -424,7 +426,48 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+If an optional `mapped-databases` is used that filter is applied first and the renaming is applied after.
+NOTE: mapping names adds an extra layer of abstraction and we advise to use this as a temporary migration solution only. It becomes harder to debug where a virtual (remapped) table actually is coming from.
```
massdosage, line 469: I'd move this to the top of the section to highlight this isn't something we recommend as early as possible.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -59,15 +64,20 @@ private CloseableThriftHiveMetastoreIface createClient(AbstractMetaStore metaSto
...
   }
 
+  @SuppressWarnings("resource")
   @Override
   public MetaStoreMapping newInstance(AbstractMetaStore metaStore) {
     LOG
         .info("Mapping databases with name '{}' to metastore: {}", metaStore.getName(),
             metaStore.getRemoteMetaStoreUris());
-    MetaStoreMapping mapping = new MetaStoreMappingImpl(prefixNameFor(metaStore), metaStore.getName(),
+    MetaStoreMappingImpl metaStoreMappingImpl = new MetaStoreMappingImpl(prefixNameFor(metaStore), metaStore.getName(),
```
massdosage, line 73: Can we no longer use the Interface type on the LHS here?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -97,9 +96,6 @@ private void add(AbstractMetaStore metaStore) {
   }
 
   private DatabaseMapping createDatabaseMapping(MetaStoreMapping metaStoreMapping) {
```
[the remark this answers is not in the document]
massdosage: Not touched or *now* touched?

Is this worth documenting in the changelog?

Should we not remove the IdentityMapping class to be clear? Or at least deprecate it?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
```diff
@@ -42,6 +44,7 @@
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.junit.MockitoJUnitRunner;
+import org.tuckey.web.filters.urlrewrite.utils.StringUtils;
```
massdosage, line 47: Is this really the `StringUtils` you want to use?

massdosage (changes requested):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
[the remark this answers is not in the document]
patduin: now, sorry yes those are all good questions. We use it in some tests I think I haven't looked that much. For users it really shouldn't make any difference in theory but yeah its a different code path so there might be some unforeseen things.

README.md
patduin: yes it doesn't and yes I tried to keep it to a minimum, it's tricky :)

README.md
patduin: you cannot use the old name e.g. hotel_booking. Only the "mapped" name is exposed. But that's because you're looking through Waggle Dance tinted glasses.

README.md
patduin: that's a good point :)

README.md
massdosage: Ah OK, so some of my suggestions might be incorrect then, please ignore/update them accordingly.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
patduin: we can, I'll change it

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
patduin: lol

README.md
```diff
@@ -424,7 +426,49 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+Waggle Dance allows one to refer to databases by names that are different to what they are defined in the Hive metastore via a `database-name-mapping` configuration. This feature can be useful when migrating data from existing databases into different environments.
+To clarify how this feature could be used, below is an example use case:
+As part of a data migration we have decided that we want to store all Hive tables related to hotel bookings in a database called `booking`. However we have legacy data lakes that have tables in this domain but they are stored in databased with names that don't match this new standard, e.g. 'datawarehouse' or 'hotel_booking'. When federating queries to those legacy data lakes it helps data discovery if the database names are standardized to new name. Renaming a database in Hive is easy but updating all the relevant ETL scripts and jobs might not be, so to make migration easier you can configure Waggle Dance to map the database names so both the new and old names can be used. At some point in the future when all the references to the old names are updated this mapping can then be removed.
```
massdosage, line 435: Now that we've discussed this more I see my use case below isn't correct. This feature is more about allowing you to refer to things by the name they _will_ have in the future so you can future proof your ETL and scripts etc. and then migrate them later so you build up less technical debt while a migration is in progress. @barnharts4 if I'm correct would you mind having a go at using your wordsmithery and writing a better use case?

barnharts4: Actually, I was under the impression that the use-case *was* for referring to databases by their old name for federation so as to allow a smooth migration for legacy federation customers.  Then, when the customer has had time to migrate all ETL processes and scripts to the new name, the old name in the Waggledance config could be deprecated.  That is certainly the use-case we had in mind for our internal EGDP migration.

patduin: so you would want both names? the federated one and as an extra a renamed one?

barnharts4: Yes, that would be what is most helpful for doing a migration of database names with minimal disruption to legacy customers.

patduin: ok will revisit this, thanks

patduin: @barnharts4 @massdosage Ok updated the PR please have a look if you got time and let me if this makes more sense :) Thanks!

README.md
```diff
@@ -424,7 +426,51 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
+
+    federated-meta-stores:
+      - remote-meta-store-uris: thrift://10.0.0.1:9083
+        name: y
+        mapped-databases:
+        - datawarehouse
+        database-name-mapping:
+          datawarehouse: booking
+
+Note: Both the 'datawarehouse' and the mapped name 'booking' is show in X, so the mapping adds an additional virtual database mapping both to the same remote database. You can only map one extra name and you cannot map different databases to the same name.
```
barnharts4, line 459:
```suggestion
Note: Both the 'datawarehouse' name and the mapped name 'booking' are shown in X, so the mapping adds an additional virtual database mapping to the same remote database. You can only map one extra name and you cannot map different databases to the same name.
```

barnharts4: Just out of curiosity, is the restriction `You can only map one extra name` a technical restriction, or just something that was picked as a limit?  It seems from your other changed code using `transformOutboundDatabaseNameMultiple` that it returns a List, which would imply any number of mappings for a database.  But I freely admit I don't understand the internals of Waggledance.

patduin: it's me being a little lazy and a technical thingy, so in yaml it's a map so I'd have to make this into a list of mappings. It can be done wasn't sure if that's needed. Happy to change it around though :)

barnharts4: We don't have a use-case for that, so it sounds like too much trouble for something with little utility.  Was more just curious.  Thanks!

README.md
```diff
@@ -149,13 +149,15 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `primary-meta-store.latency`                            | No       | Indicates the acceptable slowness of the metastore in **milliseconds** for increasing the default connection timeout. Default latency is `0` and should be changed if the metastore is particularly slow. If you get an error saying that results were omitted because the metastore was slow, consider changing the latency to a higher number.|
 | `primary-meta-store.mapped-databases`                   | No       | List of databases to federate from the primary metastore; all other databases will be ignored. This property supports both full database names and [Java RegEx patterns](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) (both being case-insensitive). By default, all databases from the metastore are federated. |
+| `primary-meta-stores.database-name-mapping`             | No       | BiDirectional Map of database names and mapped names. key=`<database name as known in the primary metastore>`, value=`<name that should be shown to a client>`. See the [Database Name Mapping](#database-name-mapping) section.|
```
massdosage, line 152:
```suggestion
| `primary-meta-stores.database-name-mapping`             | No       | BiDirectional Map of database names and mapped name, where key=`<database name as known in the primary metastore>` and value=`<name that should be shown to a client>`. See the [Database Name Mapping](#database-name-mapping) section.|
```

README.md
```diff
@@ -149,13 +149,15 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `federated-meta-stores[n].latency`                      | No       | Indicates the acceptable slowness of the metastore in **milliseconds** for increasing the default connection timeout. Default latency is `0` and should be changed if the metastore is particularly slow. If you get an error saying that results were omitted because the metastore was slow, consider changing the latency to a higher number.|
 | `federated-meta-stores[n].mapped-databases`             | No       | List of databases to federate from this federated metastore, all other databases will be ignored. This property supports both full database names and [Java RegEx patterns](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) (both being case-insensitive). By default, all databases from the metastore are federated. |
+| `federated-meta-stores[n].database-name-mapping`             | No       | BiDirectional Map of database names and mapped names. key=`<database name as known in the federated metastore>`, value=`<name that should be shown to a client>`. See the [Database Name Mapping](#database-name-mapping) section.|
```
massdosage, line 160:
```suggestion
| `federated-meta-stores[n].database-name-mapping`             | No       | BiDirectional Map of database names and mapped names where key=`<database name as known in the federated metastore>` and value=`<name that should be shown to a client>`. See the [Database Name Mapping](#database-name-mapping) section.|
```

README.md
```diff
@@ -424,7 +426,51 @@ Assumes database resolution is done by adding prefixes. If database resolution i
...
-    
+
+## Database Name Mapping
+NOTE: mapping names adds an extra layer of abstraction and we advise to use this as a temporary migration solution only. It becomes harder to debug where a virtual (remapped) table actually is coming from.
+
+Waggle Dance allows one to refer to databases by names that are different to what they are defined in the Hive metastore via a `database-name-mapping` configuration. This feature can be useful when migrating data from existing databases into different environments.
+To clarify how this feature could be used, below is an example use case:
+As part of a data migration we have decided that we want to store all Hive tables related to hotel bookings in a database called `booking`. However we have legacy data lakes that have tables in this domain but they are stored in databases with names that don't match this new standard, e.g. 'datawarehouse'. To ease migration we want to expose the 'datawarehouse' db as both the original name and the booking name. This way consumers can start using the new name while producers migrate their scripts from the old name or visa versa. When the migration is done the mapping can be removed and the database renamed.
```
massdosage, line 435:
```suggestion
As part of a data migration we have decided that we want to store all Hive tables related to hotel bookings in a database called `booking`. However we have legacy data lakes that contain booking-related tables which are stored in databases with other names e.g. 'datawarehouse'. To ease migration we want to expose the 'datawarehouse' database via both the original name and the `booking` name. This way consumers can start using the new name while producers migrate their scripts from the old name or vice versa. When the migration is done the mapping can be removed and the database renamed.
```

massdosage (changes requested):

massdosage (approved):

README.md
```diff
@@ -424,7 +426,51 @@ Assumes database resolution is done by adding prefixes. If database resolution i
      where h.date = '2016-05-13'
        and h.hour = 1
     ;
-    
+
+## Database Name Mapping
+NOTE: mapping names adds an extra layer of abstraction and we advise to use this as a temporary migration solution only. It becomes harder to debug where a virtual (remapped) table actually is coming from.
```
jmnunezizu, line 431: Nice.

abhimanyugupta07 (approved):

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
```diff
@@ -161,6 +164,23 @@ public void setMappedDatabases(List<String> mappedDatabases) {
     this.mappedDatabases = mappedDatabases;
   }
 
+  public Map<String, String> getDatabaseNameMapping() {
+    return databaseNameMapping;
+  }
+
+  public void setDatabaseNameMapping(Map<String, String> databaseNameMapping) {
+    if (databaseNameMapping == null) {
```
jmnunezizu, line 172: Would it be better to initialise the databaseNameMapping at the class property level and avoid the initialisation in this setter?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/PrefixMapping.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.mapping.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PrefixMapping extends MetaStoreMappingDecorator {
```
jmnunezizu, line 21: It seems to me that this is the actual decorator and that `MetastoreMappingDecorator` is just delegating the calls but not decorating anything.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseNameMapping.java
```diff
@@ -0,0 +1,73 @@
...
+      outbound = new HashMap<>(databaseNameMap);
+    } else {
+      inbound = Collections.emptyMap();
+      outbound = Collections.emptyMap();
+    }
+  }
+
+  @Override
+  public String transformOutboundDatabaseName(String databaseName) {
+    return transformOutboundDatabaseNameMultiple(databaseName).get(0);
```
jmnunezizu, line 49: Is it possible to get an empty `List`? That could throw an `IndexOutOfBoundsException`.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/PrefixMapping.java
patduin: yeah I think that's how you write it though you have an abstract "decorator" class and then concrete subclasses. But my head first java design patterns has been a while :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseNameMapping.java
patduin: no

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
patduin: if I do it at the class level then it can get overwritten by the setter I just never want this to be null

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/PrefixMapping.java
jmnunezizu: Yes, your implementation is spot on and just how it's recommended in the GoF book. Having the abstract class is useful if you've got many decorators. If you don't, you can just simplify it and have a single implementation that decorates.

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
jmnunezizu: Right, I see.

I don't know if calling the getter before the setter is a valid use case. The class would be in an illegal state and you'd get a `null` rather than an empty map.

I think I'd probably change the setter to throw an exception if a `null` value is provided, and instantiate the collection at the property level. It gives a much better API in my opinion.

It's fine if you leave it like this too :)

jmnunezizu (approved):

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
patduin: How about I change the getter to return empty if null. then we cover that case as well

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/PrefixMapping.java
patduin: we have 2 concrete subclasses

jmnunezizu: I missed the other one then. Thanks.

PR #201, opened 2020-11-18
Feature/Mapped tables

andreeapad:
Implements mapped tables feature: #195


waggle-dance-api/src/test/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStoreTest.java
```diff
@@ -173,6 +174,72 @@ public void emptyMappedDatabases() {
...
+    MappedTables mappedTables = new MappedTables("valid_db", null);
+    metaStore.setMappedTables(Lists.newArrayList(mappedTables));
+
+    Set<ConstraintViolation<T>> violations = validator.validate(metaStore);
+    assertThat(violations.size(), is(1));
+  }
+
+  @Test
+  public void mappedTablesEmptyTblsInvalid() {
+    MappedTables mappedTables = new MappedTables("valid_db", Lists.newArrayList() );
```
nvitucci, line 218:
```suggestion
    MappedTables mappedTables = new MappedTables("valid_db", Lists.newArrayList());
```

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MappedTables.java
```diff
@@ -0,0 +1,50 @@
...
+  public MappedTables(String database, List<String> mappedTables) {
+    this.database = database;
+    this.mappedTables = mappedTables;
+  }
+
+  public String getDatabase() {
+    return database;
+  }
+
+  public void setDatabase(String database) {
```
nvitucci, line 39: Are the setter methods useful here? Shouldn't the fields be immutable?

andreeapad: I think the setter methods are used by the yaml parser in conjunction with the no args constructor. I'm not entirely sure that's how it works though. The `MetastoreTunnel` class seems to have the same.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -93,6 +97,16 @@ private void add(AbstractMetaStore metaStore) {
     mappingsByPrefix.put(metaStoreMapping.getDatabasePrefix(), databaseMapping);
     Whitelist mappedDbWhitelist = new Whitelist(metaStore.getMappedDatabases());
     mappedDbByPrefix.put(metaStoreMapping.getDatabasePrefix(), mappedDbWhitelist);
+
+    List<MappedTables> mappedTables = metaStore.getMappedTables();
+    if (mappedTables != null) {
+    Map<String, Whitelist> mappedTblByDb = new ConcurrentHashMap<>();
```
JayGreeeen, line 103: Formatting looks off here

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -202,6 +216,42 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSu
...
+    }
+  }
+
+  @Override
+  public List<String> filterTables(String databaseName, List<String> tableNames, DatabaseMapping mapping) {
+    List<String> filteredTables = new ArrayList<>();
+    String databasePrefix = mapping.getDatabasePrefix();
+    String transformedDb = mapping.transformInboundDatabaseName(databaseName);
+    for (String table: tableNames)
+    if (isTableWhitelisted(databasePrefix, transformedDb, table)) {
```
JayGreeeen, line 235: Formatting here too. Also no brackets in for loop?

andreeapad: yeah, I don't know what happened

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/YamlFederatedMetaStoreStorageTest.java
```diff
@@ -161,6 +162,36 @@ public void loadFederationMultipleFederations() throws Exception {
     assertThat(storage.getAll().get(2), is(metaStore));
   }
 
+  @Test
+  public void loadFederationMappedDatabasesAndTables() throws Exception {
+    File f = dataFolder.getFile("mapped-tables.yml");
```
patduin, line 167: f=file

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/YamlFederatedMetaStoreStorageTest.java
```diff
@@ -161,6 +162,36 @@ public void loadFederationMultipleFederations() throws Exception {
...
+    assertThat(federatedMetastore.getMappedDatabases(), is(Lists.newArrayList("db1", "db2")));
+    assertThat(federatedMetastore.getMappedTables().get(0).getDatabase(), is("db1"));
+    assertThat(federatedMetastore.getMappedTables().get(0).getMappedTables(), is(Lists.newArrayList("tbl1")));
+    assertThat(federatedMetastore.getMappedTables().get(1).getDatabase(), is("db2"));
+    assertThat(federatedMetastore.getMappedTables().get(1).getMappedTables(), is(Lists.newArrayList("tbl2")));
+  }
+
+  @Test(expected = ConstraintViolationException.class)
+  public void loadFederationMappedTablesEmptyTablesInvalid() throws Exception {
+    File f = dataFolder.getFile("mapped-tables-empty-tables.yml");
```
patduin, line 190: f=file

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/YamlFederatedMetaStoreStorageTest.java
andreeapad: Ok, I'll change it in the previous tests too

patduin (approved):
this looks really good

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/server/FederatedHMSHandlerTest.java
```diff
@@ -345,17 +345,20 @@ public void get_tables() throws TException {
     when(primaryMapping.transformInboundDatabaseName(DB_P)).thenReturn("inbound");
     List<String> tables = Lists.newArrayList("table1");
     when(primaryClient.get_tables("inbound", "*")).thenReturn(tables);
+    when(databaseMappingService.filterTables(DB_P, tables, primaryMapping)).thenReturn(tables);
     List<String> result = handler.get_tables(DB_P, "*");
     assertThat(result, is(tables));
   }
 
   @Test
   public void get_all_tables() throws TException {
```
nvitucci, line 354: Any idea why the method names here are snake case?

waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/MappedTables.java
nvitucci: Ok, if removing them can cause troubles, we can leave them.

nvitucci (approved):
Looks good.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/server/FederatedHMSHandlerTest.java
andreeapad: I think because those are their names in the original HMSHandler and it's easier to just copy them.

patduin: yes

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -60,7 +60,7 @@ private CloseableThriftHiveMetastoreIface createClient(AbstractMetaStore metaSto
       return metaStoreClientFactory.newInstance(metaStore);
     } catch (Exception e) {
       LOG.error("Can't create a client for metastore '{}':", metaStore.getName(), e);
-      return newUnreachableMetatstoreClient(metaStore);
+      return newUnreachableMetastoreClient(metaStore);
```
massdosage, line 63: Wow, well spottted!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -93,6 +97,16 @@ private void add(AbstractMetaStore metaStore) {
     mappingsByPrefix.put(metaStoreMapping.getDatabasePrefix(), databaseMapping);
     Whitelist mappedDbWhitelist = new Whitelist(metaStore.getMappedDatabases());
     mappedDbByPrefix.put(metaStoreMapping.getDatabasePrefix(), mappedDbWhitelist);
+
+    List<MappedTables> mappedTables = metaStore.getMappedTables();
+    if (mappedTables != null) {
+      Map<String, Whitelist> mappedTblByDb = new ConcurrentHashMap<>();
```
massdosage, line 103: Why does this need to be a `ConcurrentHashMap` instead of just a normal `HashMap`?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -93,6 +97,16 @@ private void add(AbstractMetaStore metaStore) {
     mappingsByPrefix.put(metaStoreMapping.getDatabasePrefix(), databaseMapping);
     Whitelist mappedDbWhitelist = new Whitelist(metaStore.getMappedDatabases());
     mappedDbByPrefix.put(metaStoreMapping.getDatabasePrefix(), mappedDbWhitelist);
+
+    List<MappedTables> mappedTables = metaStore.getMappedTables();
+    if (mappedTables != null) {
```
massdosage, line 102: Could this ever be an empty list? If so would it be OK to add an empty map on line 108?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingService.java
```diff
@@ -37,6 +37,11 @@
    */
   DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSuchObjectException;
 
+  void checkTable(String databaseName, String tableName,
+      DatabaseMapping mapping) throws NoSuchObjectException;
+
+  List<String> filterTables(String databaseName, List<String> tableNames, DatabaseMapping mapping);
```
massdosage, line 43: This could do with some javadoc explaining what the intention is, I got a bit confused reading the implementations below. Is this returning a list of table names that should be filtered out? Or a list of table names that don't pass the filter? What are the criteria for the "filtering" operation?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -202,6 +216,43 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSu
...
+  }
+
+  @Override
+  public List<String> filterTables(String databaseName, List<String> tableNames, DatabaseMapping mapping) {
+    List<String> filteredTables = new ArrayList<>();
+    String databasePrefix = mapping.getDatabasePrefix();
+    String transformedDb = mapping.transformInboundDatabaseName(databaseName);
+    for (String table: tableNames) {
+      if (isTableWhitelisted(databasePrefix, transformedDb, table)) {
+        filteredTables.add(table);
```
massdosage, line 236: See my comment on the javadoc on the interface above, I'm a bit confused but maybe it's because I haven't had my morning coffee yet ;) This looks like it's adding a table that is whitelisted to something that needs to be filtered out but I think the intention is actually the opposite and this is returning tables that are "vaild"? Perhaps there's some confusion (or I'm misunderstanding) what "filter" means in this context.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/Whitelist.java
```diff
@@ -29,18 +29,18 @@
 
   public Whitelist() {}
```
massdosage, line 30: Seeing as you've looked at this class as part of this, what would the impact be of renaming this to `AllowList`? Would this be invisible to end users or does the YAML somehow refer to it? If we can rename this without any end user impact I think we should. We can also do that in a separate PR.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingServiceTest.java
```diff
@@ -275,6 +271,65 @@ public void close() throws IOException {
...
+  @Test
+  public void filterTables() throws NoSuchObjectException {
+    List<String> allowedTables = Lists.newArrayList("table", "another_table");
+    primaryMetastore.setMappedDatabases(Collections.singletonList(PRIMARY_DB));
+    MappedTables mappedTables = new MappedTables(PRIMARY_DB, allowedTables);
+    primaryMetastore.setMappedTables(Collections.singletonList(mappedTables));
+    service = new PrefixBasedDatabaseMappingService(metaStoreMappingFactory,
+        Arrays.asList(primaryMetastore, federatedMetastore), queryMapping);
+    DatabaseMapping mapping = service.databaseMapping(PRIMARY_DB);
+    List<String> result = service.filterTables(PRIMARY_DB,
```
massdosage, line 327: Ah, OK, this now makes it clear to me what `filterTables` does, it filters out those that don't pass the allow list. I think just adding javadoc on the interface to make it a bit clearer is all that's needed here to address my concerns.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/YamlFederatedMetaStoreStorageTest.java
```diff
@@ -62,24 +63,24 @@ public void init() {
 
   @Test
   public void loadFederation_empty() throws Exception {
-    File f = dataFolder.getFile("empty-file.yml");
-    YamlFederatedMetaStoreStorage storage = new YamlFederatedMetaStoreStorage(f.toURI().toString(), configuration);
+    File file = dataFolder.getFile("empty-file.yml");
```
massdosage, line 66: Thank you!

massdosage (changes requested):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/Whitelist.java
andreeapad: It's invisible for the users so I'll change it.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingService.java
andreeapad: yeah... I'll add a javadoc 👍

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: I guess I could rename `filteredTables` to `allowedTables`?

massdosage: Yeah, if I understand it correctly, what it's doing is `filterOutUnmappedTables` but that doesn't read very nicely ;) Let's try your suggestion and see what that looks like along with the javadoc.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Hmm, I looked at the other maps in the class and I assumed they all have to be concurrent/synchronised, but since we are not making any changes to `mappedTblByDb` after initialisation, it should be ok to use a normal `HashMap`.

andreeapad: I was looking at the `mappedTblByPrefix` above actually, and I think that one doesn't have to be a `ConcurrentHashMap` either?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Yes, this could be an empty list and in that case `mappedTblByDb` stays empty and is added on line 108, so I think that's already happening.

massdosage: OK, as long as that's not going to cause any issues then I guess it's fine! Worth checking this path happens in at least one of the tests.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
massdosage: I think the reason that `mappedDbByPrefix` is a `ConcurrentHashMap` applies to `mappedTblByPrefix` as that's a member variable of the class and the `add` method could be called by multiple threads which in turn could update the map concurrently. This doesn't apply to `mappedTblByDb` as it's a method level variable so will be instantiated separately for each caller.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/Whitelist.java
andreeapad: Should all the places which mention `whitelist` be changed to `allowList` too? Like `databaseToTableWhitelist` -> `databaseToTableAllowList`

massdosage: If this can all be done in a way that doesn't affect end users then yes please.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Alright, that makes sense

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Ok, added 1 test

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/DatabaseMappingService.java
```diff
@@ -37,6 +37,14 @@
    */
   DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSuchObjectException;
 
+  void checkTable(String databaseName, String tableName,
```
massdosage, line 40: This could also do with some javadoc explaining what exactly it's checking and in what case the `NoSuchObjectException` will be thrown.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -202,6 +217,43 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSu
         "Waggle Dance error no database mapping available tried to map database '" + databaseName + "'");
   }
 
+  @Override
+  public void checkTable(String databaseName, String tableName, DatabaseMapping mapping)
```
massdosage, line 221: I wonder if `isTableAllowed` would be a better name for this? `check` feels a bit vague to me.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -202,6 +217,43 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSu
...
+      throws NoSuchObjectException {
+    String databasePrefix = mapping.getDatabasePrefix();
+    String transformedDbName = mapping.transformInboundDatabaseName(databaseName);
+    if (!isTableAllowed(databasePrefix, transformedDbName, tableName)) {
+      throw new NoSuchObjectException(String.format("%s.%s table not found", databaseName, tableName));
```
massdosage, line 226:
```suggestion
      throw new NoSuchObjectException(String.format("%s.%s table not found in any mappings", databaseName, tableName));
```
^ just to be clear that it's not a table not found in the metastore.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
massdosage: or `checkTableAllowed` since it doesn't return a `boolean`.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -291,6 +305,34 @@ public DatabaseMapping databaseMapping(@NotNull String databaseName) throws NoSu
     throw new NoSuchObjectException("Primary metastore does not have database " + databaseName);
   }
 
+  @Override
+  public void checkTable(String databaseName, String tableName,
+      DatabaseMapping mapping) throws NoSuchObjectException {
+    if (!isTableAllowed(databaseName, tableName)) {
+      throw new NoSuchObjectException(String.format("%s.%s table not found", databaseName, tableName));
```
massdosage, line 312:
```suggestion
      throw new NoSuchObjectException(String.format("%s.%s table not found in any mappings", databaseName, tableName));
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: hmm yeah that's maybe a good point we want to indicate is because of WaggleDance probably. This reminds me @andreeapad what happens if you want to create a table that isn't mapped? Does it get created and then not shown or is creation not allowed?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -176,6 +176,24 @@ private DatabaseMapping checkWritePermissions(String databaseName) throws TExcep
     return mapping;
   }
 
+  private DatabaseMapping getDbMappingAndCheckTable(String dbName, String tblName) throws NoSuchObjectException {
```
massdosage, line 179:
```suggestion
  private DatabaseMapping getDbMappingAndCheckTableAllowed(String dbName, String tblName) throws NoSuchObjectException {
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Ok, I'm not sure whether we want to have the same message that Hive would give when the table doesn't exist.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -176,6 +176,24 @@ private DatabaseMapping checkWritePermissions(String databaseName) throws TExcep
...
+    DatabaseMapping mapping = databaseMappingService.databaseMapping(dbName);
+    databaseMappingService.checkTable(dbName, tblName, mapping);
+    return mapping;
+  }
+
+  private DatabaseMapping checkWritePermissionsAndCheckTable(String dbName, String tblName) throws TException {
```
massdosage, line 185:
```suggestion
  private DatabaseMapping checkWritePermissionsAndTableAllowed(String dbName, String tblName) throws TException {
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Not really, cause it only throws an error, it doesn't return any boolean.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -176,6 +176,24 @@ private DatabaseMapping checkWritePermissions(String databaseName) throws TExcep
...
+    DatabaseMapping mapping = checkWritePermissions(dbName);
+    databaseMappingService.checkTable(dbName, tblName, mapping);
+    return mapping;
+  }
+
+  private void checkWritePermissionsAndCheckTable(String dest_db, String dest_table_name, DatabaseMapping mapping)
```
massdosage, line 191:
```suggestion
  private void checkWritePermissionsAndCheckTableAllowed(String destDatatabaseName, String destTableName, DatabaseMapping mapping)
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: ah right, the second comment appeared later 😄

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: I think the same exception is good but a different message so users at least have some hint that this is possible to change

andreeapad: @patduin  it should get created then not shown

massdosage (changes requested):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Ok, so I'll change the message. The new comments appeared only after I sent mine oops

massdosage: > Ok, so I'll change the message. The new comments appeared only after I sent mine oops

That's OK, we're all comfortable with eventual consistency here :)

patduin: on the created: ok I don't there's an ideal solution here, should probably only happen in theory, generally only federated metastores will be mapped and they are mostly read only

README.md
```diff
@@ -172,6 +182,14 @@ The table below describes the metastore tunnel configuration values:
 | `*.metastore-tunnel.timeout`                            | No       | The SSH session timeout in milliseconds, `0` means no timeout. Default is `60000` milliseconds, i.e. 1 minute. |
 | `*.metastore-tunnel.strict-host-key-checking`           | No       | Whether the SSH tunnel should be created with strict host key checking. Can be set to `yes` or `no`. The default is `yes`. |
 
+The table below describes the `mapped-tables` configuration. For each entry in the list, a database name and the corresponding list of table names/patterns must be mentioned.
+
+| Property                                                | Required | Description |
+|:----|:----:|:----|
+| `*.mapped-tables[n].database`                           | Yes       | Name of the database which contains the tables to be mapped.|
+| `*.mapped-tables[n].mapped-tables`                      | Yes       | List of tables allowed for the database specified in the field above. This property supports both full table names and [Java RegEx patterns](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) (both being case-insensitive).|
```
massdosage, line 190:
```suggestion
| `*.mapped-tables[n].mapped-tables`                      | Yes       | List of tables allowed for the database specified in the field above. This property supports both full table names and [Java RegEx patterns](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) (both being case-insensitive).|
```
Minor, but we should link to the version of Java we build against.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -258,17 +310,21 @@ public PanopticOperationHandler getPanopticOperationHandler() {
       public List<TableMeta> getTableMeta(String db_patterns, String tbl_patterns, List<String> tbl_types) {
         Map<DatabaseMapping, String> databaseMappingsForPattern = databaseMappingsByDbPattern(db_patterns);
 
-        BiFunction<TableMeta, DatabaseMapping, Boolean> filter = (tableMeta, mapping) -> isWhitelisted(
-            mapping.getDatabasePrefix(), tableMeta.getDbName());
-
+        BiFunction<TableMeta, DatabaseMapping, Boolean> filter = (tableMeta, mapping) -> {
```
massdosage, line 313: Minor, but I generally think if the code on the RHS of a `->` is more than a line or two then it probably warrants being extracted into its own method. It won't make that much difference here since there's nothing else in this method anyway. Up to you whether you want to leave it like this.

massdosage (approved):
Nice work!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
andreeapad: Yeah, probably better have a separate method.

PR #202, opened 2020-11-23
Clarified log4j2 usage

nvitucci:


patduin (approved):

waggle-dance/src/main/scripts/waggle-dance.sh
```diff
@@ -11,6 +11,6 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 SCRIPT_DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
-java -jar $SCRIPT_DIR/../service/waggle-dance-core-latest-exec.jar "$@"
+java -Dlogging.config=$SCRIPT_DIR/../conf/log4j2.xml -jar $SCRIPT_DIR/../service/waggle-dance-core-latest-exec.jar "$@"
```
massdosage, line 14: Hmmm, another way do this would be to add a logging section to the README a bit like https://github.com/HotelsDotCom/circus-train/#logging (but describing log4j instead of log4j) then this can be set in the YAML. If we set it in the YAML does it then override the value set here? If so we could keep this for those who haven't set it in the YAML. Also, did it not output everything to the console by default? It used to (see last line under section https://github.com/HotelsDotCom/waggle-dance/#running-on-the-command-line)

nvitucci: > Also, did it not output everything to the console by default? It used to (see last line under section https://github.com/HotelsDotCom/waggle-dance/#running-on-the-command-line)

It was, but when we tried to change the logging level it looked like nothing changed. We started seeing debug lines only when we added this explicitly in the script.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## [3.8.0] - TBD
 ### Changed
 * Updated `hotels-oss-parent` to 6.1.0 (was 5.0.0).
-* Updated `waggle-dance.sh` script to include the default log4j2 configuration.
+* Updated README with logging configuration.
```
massdosage, line 4: Doesn't need to be mentioned in the CHANGELOG as this isn't really part of a "release".

massdosage (changes requested):

patduin (approved):

CHANGELOG.md
nvitucci: True. Removing.

patduin (approved):

massdosage (approved):

PR #205, opened 2021-01-06
Update copyrights for externally sourced code

massdosage:


pom.xml
```diff
@@ -266,15 +266,7 @@
           <configuration>
             <!-- excluding files that don't need a header update -->
             <excludes>
-              <exclude>src/main/java/com/hotels/bdp/waggledance/api/validation/constraint/TunnelRoute.java</exclude>
```
massdosage, line 269: These classes are no longer in this project so they can be removed from this exclusion section.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2015-2017 The Apache Software Foundation and Expedia Inc.
```
massdosage, line 2: I've added the missing comma(!) and updated the year.

JayGreeeen: out of interest, will this ever be updated to Expedia Group?

massdosage: According to our lawyers, in the near future, no. I'm guessing it's the name of some legally registered entity which is hard/expensive to change.

nvitucci (approved):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/MetaStoreProxyServer.java
JayGreeeen: ahh okay

JayGreeeen (approved):

PR #206, opened 2021-02-10
Feature/Support metastore filter hook

### LINKED ISSUE #77: Use configuration-properties in clients HiveConf
_I'd like to use the `configuration-properties` that I can set in the `waggle-dance-server.yml` in HiveConf that is used to create Thrift Clients._

Currently we use the properties only in the Handler (the server), it would be good if we can set properties and override settings used in the clients. 
For instance see `ThriftMetastoreClient` it contains this code:
`    retries = HiveConf.getIntVar(conf, HiveConf.ConfVars.METASTORETHRIFTCONNECTIONRETRIES);
`
But there is currently no way to override this cause we create the HiveConf ourselves in `CloseableThriftHiveMetastoreIfaceClientFactory`. Thus that values is hardcoded to the Hive default  3.

### Acceptance Criteria:

- Use the`configuration-properties`(see `WaggleDanceConfiguration` class) when creating the HiveConf for the clients.
- Documents it's use in the README (it seems to be a hidden feature now).

andreeapad:
Introduces support for metastore filter hooks and closes #77


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -66,7 +72,12 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import sun.security.krb5.RealmException;
```
massdosage: I think you need to optimise the imports here, look like leftovers from stuff you've tried and removed?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -91,6 +97,30 @@ private CloseableThriftHiveMetastoreIface newUnreachableMetastoreClient(Abstract
...
+    }
+    for (Map.Entry<String, String> property: waggleDanceConfiguration.getConfigurationProperties().entrySet()) {
+      conf.set(property.getKey(), property.getValue());
+    }
+    conf.set(HiveConf.ConfVars.METASTORE_FILTER_HOOK.varname, metaStoreFilterHook);
+//        "com.hotels.bdp.waggledance.mapping.model.AlluxioMetastoreFilter");
+    Class<? extends MetaStoreFilterHook> filterHookClass = conf.getClass(
+        HiveConf.ConfVars.METASTORE_FILTER_HOOK.varname, DefaultMetaStoreFilterHookImpl.class,
+        MetaStoreFilterHook.class);
+    String msg = "Unable to create instance of " + filterHookClass.getName() + ": ";
```
massdosage: You can move this down to line 120 no? (I understand this draft and stuff here likely to change and be tidied up).

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/AlluxioMetastoreFilter.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.hadoop.hive.metastore.api.MetaException;
+import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
+import org.apache.hadoop.hive.metastore.api.Partition;
+import org.apache.hadoop.hive.metastore.api.PartitionSpec;
+import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
+import org.apache.hadoop.hive.metastore.api.Table;
+
+public class AlluxioMetastoreFilter implements MetaStoreFilterHook {
+
+  public static final String S3_PREFIX = "s3://";
```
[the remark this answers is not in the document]
andreeapad: I have just updated this to only be used in the tests (not sure in which package to add it though) since whoever uses this will bring their own filter and in the case of the Alluxio filter it will have to be configured with the right Alluxio path  depending on the location, so what I had here is a basic version that wouldn't be of much use.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/NullAuthorizationProvider.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.hadoop.hive.metastore.api.Database;
+import org.apache.hadoop.hive.ql.metadata.AuthorizationException;
+import org.apache.hadoop.hive.ql.metadata.HiveException;
+import org.apache.hadoop.hive.ql.metadata.Partition;
+import org.apache.hadoop.hive.ql.metadata.Table;
+import org.apache.hadoop.hive.ql.security.HiveAuthenticationProvider;
+import org.apache.hadoop.hive.ql.security.authorization.HiveAuthorizationProvider;
+import org.apache.hadoop.hive.ql.security.authorization.Privilege;
+
+public class NullAuthorizationProvider implements HiveAuthorizationProvider {
```
massdosage, line 30: What's this class needed for?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/AlluxioMetastoreFilter.java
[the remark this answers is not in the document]
massdosage: OK, which downstream modules need this? If its only one you can just move it to the src/test folder of that module. If its multiple then we'll need to see if we create some kind of a tests-artifacts that can be depended on downstream.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/NullAuthorizationProvider.java
andreeapad: I'll delete this. When I was first testing things I thought we needed to override the default authorizer, but that changed when we decided to load the filter ourselves.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
andreeapad: yes, moved it

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/AlluxioMetastoreFilter.java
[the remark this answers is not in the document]
andreeapad: Yeah, it's in 2 places, one is in the integration tests and one is in the core module, so atm I have it in core since the integration tests depend on that one anyway.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
andreeapad: yes, should be deleted now I think.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/AlluxioMetastoreFilter.java
```diff
@@ -0,0 +1,131 @@
...
+    setAlluxioLocation(table.getSd());
+  }
+
+  private void setAlluxioLocation(Partition partition) {
+    setAlluxioLocation(partition.getSd());
+  }
+
+  private void setAlluxioLocation(StorageDescriptor sd) {
+    String location = sd.getLocation();
+    sd.setLocation(ALLUXIO_PREFIX + location);
```
patduin, line 128: this is incorrect as it doesn't replace anything this will turn `s3://<path>` into `alluxio://s3://<path>`

patduin (commented):
looks good, we need some readme update on how to use and limitations.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/AlluxioMetastoreFilter.java
andreeapad: yes, it used to replace s3 with alluxio in the first version but I just kept it for testing and since in the tests we don't have s3 locations, I deleted the real logic.

patduin: ok but this class is in src/main/java can we move it to src/test in that case. If in main it should be correct. If we move it just for test we should all reference to alluxio as that's unrelated then it's just a PrefixingMetastoreFilter in that case.

andreeapad: yeah, I would have moved it to test but then it's not seen in the integration tests module. @massdosage suggested to create some test artifact, I'll try to do that

patduin: maybe it's less work if we just make it generic and configurable and leave it in here as a reference implementation??

patduin: might even be usuable

andreeapad: Configurable how? In terms of location? I moved the filter to test and added the test artifact yesterday and it works. Also, maybe we can point to the Alluxio filter when/if that will be open-sourced as a reference implementation

patduin: if it's move to test then that's fine as is. no problem!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -91,6 +97,31 @@ private CloseableThriftHiveMetastoreIface newUnreachableMetastoreClient(Abstract
...
+    conf.set(HiveConf.ConfVars.METASTORE_FILTER_HOOK.varname, metaStoreFilterHook);
+    try {
+      Class<? extends MetaStoreFilterHook> filterHookClass = conf
+          .getClass(HiveConf.ConfVars.METASTORE_FILTER_HOOK.varname, DefaultMetaStoreFilterHookImpl.class,
+              MetaStoreFilterHook.class);
+      Constructor<? extends MetaStoreFilterHook> constructor = filterHookClass.getConstructor(HiveConf.class);
+      return constructor.newInstance(conf);
+    } catch (Exception e) {
+      String msg = "Unable to create instance of " + metaStoreFilterHook + ": ";
+      throw new WaggleDanceServerException(msg + e);
```
patduin, line 121: `msg, e`

patduin (changes requested):
just had one comment

README.md
```diff
@@ -107,6 +107,7 @@ The table below describes all the available configuration values for Waggle Danc
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
 | `status-polling-delay`            | No         | Controls the delay that checks metastore availability and updates long running connections of any status change. Default is `5` (every 5 minutes). |
 | `status-polling-delay-time-unit`  | No         | Controls the delay time unit. Default is `MINUTES` . |
+| `configuration-properties`        | No         | Map of hive properties that will be added to the HiveConf used when creating the Thrift clients. |
```
patduin, line 110: Make it clear they are shared among all clients. Hopefully ok for now we can make that per Metastore in the future if needed.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
andreeapad: WaggleDanceServerException only takes 1 string as argument

massdosage: Maybe worth adding an overridden constructor to allow one to pass a cause in?

patduin (approved):

README.md
```diff
@@ -107,6 +107,7 @@ The table below describes all the available configuration values for Waggle Danc
 | `database-resolution`             | No         | Controls what type of database resolution to use. See the [Database Resolution](#database-resolution) section. Default is `MANUAL`. |
 | `status-polling-delay`            | No         | Controls the delay that checks metastore availability and updates long running connections of any status change. Default is `5` (every 5 minutes). |
 | `status-polling-delay-time-unit`  | No         | Controls the delay time unit. Default is `MINUTES` . |
+| `configuration-properties`        | No         | Map of hive properties that will be added to the HiveConf used when creating the Thrift clients (they will be shared among all the clients). |
```
massdosage, line 110:
```suggestion
| `configuration-properties`        | No         | Map of Hive properties that will be added to the HiveConf used when creating the Thrift clients (they will be shared among all the clients). |
```

README.md
```diff
@@ -158,6 +160,7 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `primary-meta-store.mapped-databases`                   | No       | List of databases to federate from the primary metastore; all other databases will be ignored. This property supports both full database names and [Java RegEx patterns](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) (both being case-insensitive). By default, all databases from the metastore are federated. |
 | `primary-meta-store.mapped-tables`                      | No       | List of mappings from databases to tables to federate from the primary metastore, similar to `mapped-databases`. By default, all tables are available. See `mapped-tables` configuration below. |
+| `primary-meta-stores.hive-metastore-filter-hook`        | No       | Name of the class which implements the `MetaStoreFilterHook` interface from Hive. This allows a metastore filter hook to be applied on the corresponding Hive metastore calls. Can be configured with the `configuration-properties` specified in the `waggle-dance-server.yml` configuration. They will be added in the HiveConf object that is given to the constructor of the `MetaStoreFilterHook` implementation you provide. |
```
massdosage, line 163:
```suggestion
| `primary-meta-stores.hive-metastore-filter-hook`        | No       | Name of the class which implements the `MetaStoreFilterHook` interface from Hive. This allows a metastore filter hook to be applied to the corresponding Hive metastore calls. Can be configured with the `configuration-properties` specified in the `waggle-dance-server.yml` configuration. They will be added in the HiveConf object that is given to the constructor of the `MetaStoreFilterHook` implementation you provide. |
```

README.md
```diff
@@ -167,9 +170,11 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `federated-meta-stores[n].mapped-databases`             | No       | List of databases to federate from this federated metastore, all other databases will be ignored. This property supports both full database names and [Java RegEx patterns](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) (both being case-insensitive). By default, all databases from the metastore are federated. |
 | `federated-meta-stores[n].mapped-tables`                | No       | List of mappings from databases to tables to federate from this federated metastore, similar to `mapped-databases`. By default, all tables are available. See `mapped-tables` configuration below. |
+| `federated-meta-stores[n].hive-metastore-filter-hook`   | No       | Name of the class which implements the `MetaStoreFilterHook` interface from Hive. This allows a metastore filter hook to be applied on the corresponding Hive metastore calls. Can be configured with the `configuration-properties` specified in the `waggle-dance-server.yml` configuration. They will be added in the HiveConf object that is given to the constructor of the `MetaStoreFilterHook` implementation you provide. |
```
massdosage, line 173:
```suggestion
| `federated-meta-stores[n].hive-metastore-filter-hook`   | No       | Name of the class which implements the `MetaStoreFilterHook` interface from Hive. This allows a metastore filter hook to be applied to the corresponding Hive metastore calls. Can be configured with the `configuration-properties` specified in the `waggle-dance-server.yml` configuration. They will be added in the HiveConf object that is given to the constructor of the `MetaStoreFilterHook` implementation you provide. |
```

pom.xml
```diff
@@ -35,6 +35,7 @@
 
   <properties>
     <maven.release.plugin.version>2.5.3</maven.release.plugin.version>
+    <maven.jar.plugin>3.2.0</maven.jar.plugin>
```
massdosage, line 38: What's wrong with the version inherited from the parent?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMapping.java
```diff
@@ -41,9 +42,9 @@
   /**
    * Outbound means parameter coming from the Hive Metastore and return result will be sent to user client.
    *
-   * @param List of databaseNames. This methods potentially returns multiple database name if configuration is setup to
-   *          map to multiple.
-   * @return
+   * @param databaseName
+   * @return List of databaseNames. This method potentially returns multiple database names if configuration is setup to
```
massdosage, line 46:
```suggestion
   * @return List of databaseNames. This method potentially returns multiple database names if configuration is set up to
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
```diff
@@ -91,6 +97,31 @@ private CloseableThriftHiveMetastoreIface newUnreachableMetastoreClient(Abstract
...
+    }
+    conf.set(HiveConf.ConfVars.METASTORE_FILTER_HOOK.varname, metaStoreFilterHook);
+    try {
+      Class<? extends MetaStoreFilterHook> filterHookClass = conf
+          .getClass(HiveConf.ConfVars.METASTORE_FILTER_HOOK.varname, DefaultMetaStoreFilterHookImpl.class,
+              MetaStoreFilterHook.class);
+      Constructor<? extends MetaStoreFilterHook> constructor = filterHookClass.getConstructor(HiveConf.class);
+      return constructor.newInstance(conf);
+    } catch (Exception e) {
+      String msg = "Unable to create instance of " + metaStoreFilterHook + ": ";
```
massdosage, line 120:
```suggestion
      String message = "Unable to create instance of " + metaStoreFilterHook;
```
The message and the stack trace will be separated in the logs anyway. You could also consider inlining this on line 121 and not having the `message` variable.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/PrefixingMetastoreFilter.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.hadoop.hive.metastore.api.Table;
+
+/**
+ * For testing purposes
+ * */
+public class PrefixingMetastoreFilter implements MetaStoreFilterHook {
+
+  public static final String PREFIX = "prefix";
+
+  public PrefixingMetastoreFilter(HiveConf conf) {
```
massdosage, line 38: Why does this need a `HiveConf` in the constructor since it's not used? Is this because of how we create it via reflection where we look for a constructor of this form?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/PrefixingMetastoreFilter.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.hadoop.hive.metastore.api.PartitionSpec;
+import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
+import org.apache.hadoop.hive.metastore.api.Table;
+
+/**
+ * For testing purposes
+ * */
+public class PrefixingMetastoreFilter implements MetaStoreFilterHook {
+
+  public static final String PREFIX = "prefix";
```
massdosage, line 36: Minor, but this might look nicer in logs or debugging if it is instead something like `prefix-` or `prefix:` otherwise it will be harder to distinguish from the path location etc.

massdosage (changes requested):

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/PrefixingMetastoreFilter.java
andreeapad: yes, it searches for that constructor

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/PrefixingMetastoreFilter.java
andreeapad: yeah, true

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/MetaStoreMappingFactoryImpl.java
andreeapad: True, I just copied the `msg` thing straight from the Hive code and I didn't pay much attention to it

pom.xml
andreeapad: Ah, I thought it wasn't specified in the parent. I just found that it's `3.1.2` so I'll delete this.

massdosage (approved):

PR #209, opened 2021-03-04
Update codeql-analysis.yml

massdosage:
The CodeQL GitHub action is failing as its decided to start using Java 11 instead of 8 (probably due to some change in `ubuntu-latest`) and WaggleDance doesn't build on Java 11. So this change sets the Java version to 8.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   CodeQL-Build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -19,11 +19,11 @@ jobs:
         # a pull request then we can checkout the head.
         fetch-depth: 2
 
-    # If this run was triggered by a pull request event, then checkout
```
massdosage, line 22: This has nothing to do with the Java version but I noticed this warning in the build while checking this so I figured I would add it:

>  issue was detected with this workflow: git checkout HEAD^2 is no longer necessary. Please remove this step as Code Scanning recommends analyzing the merge commit for best results.

patduin (approved):

JayGreeeen (approved):

PR #211, opened 2021-03-04
skip presto view parsing

patduin:


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
```diff
@@ -39,11 +39,16 @@
...
   INSTANCE;
 
+  private static final String PRESTO_VIEW_MARKER = "/* Presto View";
   private final static String RE_WORD_BOUNDARY = "\\b";
   private final static Comparator<CommonToken> ON_START_INDEX = Comparator.comparingInt(CommonToken::getStartIndex);
 
   @Override
   public String transformOutboundDatabaseName(MetaStoreMapping metaStoreMapping, String query) {
+    if (hasMarker(query)) {
+      // skipping query that are not "Hive" view queries. We can't parse those.
```
andreeapad, line 49:
```suggestion
      // skipping queries that are not "Hive" view queries. We can't parse those.
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [3.9.1] - TBD
 ### Fixed
 * Null pointer exception when creating a metastore tunnel by adding a check for null `configuration-properties`.
+* Fixing issue where Presto view are cannot be parsed resulting in errors.
```
andreeapad, line 4:
```suggestion
* Fixing issue where Presto views cannot be parsed resulting in errors.
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -88,7 +88,15 @@ public MetaStoreFilterHook getMetastoreFilter() {
 
   @Override
   public Table transformOutboundTable(Table table) {
-    table.setDbName(metaStoreMapping.transformOutboundDatabaseName(table.getDbName()));
+    String originalDatabaseName = table.getDbName();
+    String databaseName = metaStoreMapping.transformOutboundDatabaseName(originalDatabaseName);
+    table.setDbName(databaseName);
+    if (databaseName.equalsIgnoreCase(originalDatabaseName)) {
+      // Skip all the view parsing if nothing is going to change, the parsing is not without problems and we can't catch
```
andreeapad, line 95: is it worth keeping the code on line 123 since that will throw an error still?

andreeapad: the original can't have the same string as the extended cause it also needs the colon. if we add it will it still throw an error?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMappingTest.java
```diff
@@ -38,8 +38,27 @@
...
+        new DefaultMetaStoreFilterHookImpl(new HiveConf())));
+  }
+
+  @Test
+  public void transformOutboundDatabaseNamePrestoMarker() {
+    ASTQueryMapping queryMapping = ASTQueryMapping.INSTANCE;
+
+    String query = "/* Presto View */";
+
+    assertThat(queryMapping.transformOutboundDatabaseName(metaStoreMapping, query), is("/* Presto View */"));
```
massdosage, line 51:
```suggestion
    assertThat(queryMapping.transformOutboundDatabaseName(metaStoreMapping, query), is(query));
```

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMappingTest.java
```diff
@@ -38,8 +38,27 @@
...
+  }
+
+  @Test
+  public void transformOutboundDatabaseNamePrestoExpandedTextMarker() {
+    ASTQueryMapping queryMapping = ASTQueryMapping.INSTANCE;
+
+    String query = "/* Presto View: <base64 of view sql> */";
+
+    assertThat(queryMapping.transformOutboundDatabaseName(metaStoreMapping, query),
+        is("/* Presto View: <base64 of view sql> */"));
```
massdosage, line 61:
```suggestion
        is(query));
```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
```diff
@@ -88,7 +88,15 @@ public MetaStoreFilterHook getMetastoreFilter() {
 
   @Override
   public Table transformOutboundTable(Table table) {
-    table.setDbName(metaStoreMapping.transformOutboundDatabaseName(table.getDbName()));
+    String originalDatabaseName = table.getDbName();
+    String databaseName = metaStoreMapping.transformOutboundDatabaseName(originalDatabaseName);
+    table.setDbName(databaseName);
+    if (databaseName.equalsIgnoreCase(originalDatabaseName)) {
```
massdosage, line 94: I don't understand what's going on here - why is the db name the same in this case?

massdosage (changes requested):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
patduin: I don't know if it is that's why I'm checking:
if our transform didn't change the databasename we don't need to try and do the parsing just return the table as is (nothing would have changed).

This creeped in when we removed the IdentityMapping (which was just returning table for non prefixed metastores). So the if will mimic that behaviour again.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
patduin: I return the table in this if so we'll not even try to parse.

andreeapad: yes, but if the database name is different it will still throw an error

patduin: No because I skip parsing in the parser class for Presto markers. The if in here is to ensure other crazy markers don't also through errors. The code here is to basically get back to same support as in pre WD 3.6.0. WD 3.7.0 introduced the problems.
If we really have to support Presto I got this follow up: https://github.com/HotelsDotCom/waggle-dance/issues/212

andreeapad (approved):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/ASTQueryMapping.java
```diff
@@ -56,6 +61,13 @@ public String transformOutboundDatabaseName(MetaStoreMapping metaStoreMapping, S
     return result.toString();
   }
 
+  boolean hasMarker(String query) {
```
barnharts4, line 64: The function name `hasMarker` isn't super descriptive.  Can we change it to something like `hasIncompatibleViewMarker`, or `hasPrestoViewMarker`, or `hasNonHiveViewMarker`, or something like that?

patduin: sure I like `hasNonHiveViewMarker`

barnharts4 (approved):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
massdosage: OK, thanks for creating that issue. So for now we'll return the Presto view "as is" but that could mean it won't actually work since it doesn't have the correct prefixes and that's what #212 is all about.

massdosage (approved):

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java
patduin: correct this is fixing and reverting old behaviour. Make it work again if there are no prefixes to remove.

PR #215, opened 2021-03-12
fixed npe

patduin:


waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java
```diff
@@ -1034,4 +1037,27 @@ public void hiveMetastoreFilterHookConfiguredForPrimary() throws Exception {
...
+        .federate(SECONDARY_METASTORE_NAME, remoteServer.getThriftConnectionUri(), REMOTE_DATABASE)
+        .build();
+
+    runWaggleDance(runner);
+    HiveMetaStoreClient proxy = getWaggleDanceClient();
+
+    HiveObjectType objectType = HiveObjectType.DATABASE;
+    String dbName = LOCAL_DATABASE;
+    // Explicitly set to null as this threw errors
+    String objectName = null;
```
barnharts4, line 1055: Maybe a dumb question:  All the changed code looks like it is to deal with a `null` database name.  But this test sets `dbName` to `local_database` and `objectName` to `null`.   So what am I missing about what this test is testing?

patduin: The object type is checked and both as passed in. See: https://github.com/HotelsDotCom/waggle-dance/blob/main/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/model/DatabaseMappingImpl.java#L167
Which is called from here:
https://github.com/HotelsDotCom/waggle-dance/blob/main/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java#L1203

So either can be null I think but the log we got showed this:
```#get_privilege_set('HiveObjectRef(objectType:DATABASE, dbName:ean_use..7..Name:null, partValues:null, columnName:null)```

ObjectType==DATABASE
dbName not null
...7...Name truncated args but I got the same error when I set objectName to null

javsanbel2 (approved):

abhimanyugupta07 (approved):

PR #239, opened 2022-04-13
caching set_ugi so call will only go to needed metastore

patduin:
~~Still testing this in the wild.~~

The idea is to cache the `set_ugi` call in our client. Then when a "real" call is made like getTable and the connection is opened we do the delayed `set_ugi` call to only that metastore. This should prevent all those unnecessary `set_ugi` calls to metastores that won't actually be queried,
Need to test some more to make sure it works correctly. Making some assumptions on when "normally" the set_ugi is called so need to verify that's correct.

Been testing this and it looks ok. 
Hive-cli/Spark-sql connects to WD and calls setugi.
Then when a "select..." query or somesuch is made we get the "getTable/getPartition" calls with the correct set_ugi.

Ready for review

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2019 Expedia, Inc.
+ * Copyright (C) 2016-2022 Expedia, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.util.List;
 
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.thrift.transport.TTransportException;
@@ -26,6 +27,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
 
 import com.hotels.bdp.waggledance.client.compatibility.HiveCompatibleThriftHiveMetastoreIfaceFactory;
 import com.hotels.hcommon.hive.metastore.exception.MetastoreUnavailableException;
@@ -41,6 +43,8 @@ public class DefaultMetaStoreClientFactory implements MetaStoreClientFactory {
     private final String name;
     private final int maxRetries;
 
+    private HiveUgiArgs cachedUgi;
+
     private ReconnectingMetastoreClientInvocationHandler(
         String name,
         int maxRetries,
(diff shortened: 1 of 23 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -1266,14 +1274,14 @@ private DatabaseMapping checkWritePermissionsForPrivileges(PrivilegeBag privileg
   @Override
   @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
   public List<String> set_ugi(String user_name, List<String> group_names) throws MetaException, TException {
-    List<DatabaseMapping> mappings = databaseMappingService.getDatabaseMappings();
+    List<DatabaseMapping> mappings = databaseMappingService.getAllDatabaseMappings();
```
patduin, line 1277: Mostly whitespace in this class, but this is key, it gets all metastores and does the call as we will cache it and not actually call the underlying metastore we don't care if the are unavailable and it prevent availability checks.

barnharts4 (approved):
Nice!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -41,6 +43,8 @@ private static class ReconnectingMetastoreClientInvocationHandler implements Inv
     private final String name;
     private final int maxRetries;
 
+    private HiveUgiArgs cachedUgi = HiveUgiArgs.WAGGLE_DANCE_DEFAULT;
```
patduin, line 46: Setting the default so it's always set.
This can happen if Waggle Dance for instance makes a client and does call itself (via our status/health checks).
If any other Client (Trino/Presto/Qubole/Hive/Spark) might have a different lifecycle than we can trace this user back and we'll know from Metastore logs the call originated from Waggle Dance.

barnharts4: That's a great idea, love it.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
```diff
@@ -167,6 +167,8 @@ void open() {
                     + CONN_COUNT.incrementAndGet());
 
             isConnected = true;
+            LOG.info("calling #set_ugi for user '{}',  on URI {}", ugiArgs.getUser(), store);
+            client.set_ugi(ugiArgs.getUser(), ugiArgs.getGroups());
```
javsanbel2, line 171: This bit is a new addition right? I would say adding here the set helps for the use case we discussed last week

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactoryTest.java
```diff
@@ -101,7 +107,21 @@ public void defaultMethodCallThrowsTransportExceptionRetries() throws TException
 
     String result = iface.getName();
     assertThat(result, is("ourName"));
-    verify(base).reconnect();
+    verify(base).reconnect(HiveUgiArgs.WAGGLE_DANCE_DEFAULT);
+  }
+
+  @Test
+  public void set_ugi_before_call() throws Exception {
```
javsanbel2, line 114: Is this the only test related to set_ugi? If that's the case I would say we could also test when you set a custom ugi_value. We can verify the custom value get sets over the default one.

patduin: this is a custom value

javsanbel2 (approved):

PR #240, opened 2022-05-16
null and immutable list safe checks

patduin:


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/HiveUgiArgs.java
```diff
@@ -15,19 +15,23 @@
...
-  public static final HiveUgiArgs WAGGLE_DANCE_DEFAULT = new HiveUgiArgs("waggledance", Collections.emptyList());
+  public static final HiveUgiArgs WAGGLE_DANCE_DEFAULT = new HiveUgiArgs("waggledance", null);
 
   private final String user;
   private final List<String> groups;
 
   public HiveUgiArgs(String user, List<String> groups) {
     this.user = user;
-    this.groups = groups;
+    if (groups == null) {
```
patduin, line 30: I found this: https://github.com/apache/hive/blob/rel/release-2.3.9/metastore/src/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java#L6116

The I got worried that we might have an issue with this list being null or mutable so being safe for that.

jmnunezizu: You might want to log that to see if it's actually happening.

patduin: nothing calls this method and it's never null now. But yeah I will add the groups to the log on our side. It's a bit hard to tell what code gets executed on the multitude of clients and what on the server to be honest.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/ThriftMetastoreClientManager.java
```diff
@@ -167,7 +167,9 @@ void open(HiveUgiArgs ugiArgs) {
                     + CONN_COUNT.incrementAndGet());
 
             isConnected = true;
-            LOG.info("calling #set_ugi for user '{}',  on URI {}", ugiArgs.getUser(), store);
+            LOG
+                .info("calling #set_ugi for user '{}' group '{}',  on URI {}", ugiArgs.getUser(), ugiArgs.getGroups(),
```
patduin, line 171: This might get long with all the groups we'll have to see.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/HiveUgiArgs.java
patduin: on second thought I don't think we should print the groups associated to the user by default in WD that might be sensitive + there can be many groups which will make the log line unreadable.

jmnunezizu: You don't need to print the groups. You only need to print a warning saying that the list was null.

PR #241, opened 2022-05-17
Added delegation when connection is already open

patduin:


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -77,29 +77,40 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         String user = (String) args[0];
         List<String> groups = (List<String>) args[1];
         cachedUgi = new HiveUgiArgs(user, groups);
-        return Lists.newArrayList(user);
+        if (base.isOpen()) {
```
patduin, line 80: this should make sure that if setUgi is called on an open connect we pass it along. (From Hive CLI client this does not happen but perhaps other clients (Trino) it does.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -43,7 +43,7 @@ private static class ReconnectingMetastoreClientInvocationHandler implements Inv
     private final String name;
     private final int maxRetries;
 
-    private HiveUgiArgs cachedUgi = HiveUgiArgs.WAGGLE_DANCE_DEFAULT;
+    private HiveUgiArgs cachedUgi = null;
```
patduin, line 46: removed the default to be more compatible a client would behave if setUgi is never called before doing a real call. We might be seeing this in Trino clients.

javsanbel2 (approved):

barnharts4 (approved):

PR #242, opened 2022-05-19
EGDL-3124: Added Metastore availability metrics

JayGreeeen:


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/metrics/MetastoreAvailabilityMetric.java
```diff
@@ -0,0 +1,18 @@
+package com.hotels.bdp.waggledance.metrics;
+
+public enum MetastoreAvailabilityMetric {
+
+  IS_AVAILABLE(0),
+  IS_UNAVAILABLE(1),
+  IS_UNKNOWN(2);
+
+  private final int availabilityValue;
```
patduin, line 9: Can you move this to the `MetaStoreStatus` enum. Have a `getIntValue` method on it so we don't need two enums?

JayGreeeen: Didn't realise this is our thing, I will absolutely do that. Nice catch 🙌🏻

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/PollingFederationServiceTest.java
```diff
@@ -38,12 +41,15 @@
...
   private @Mock PopulateStatusFederationService populateStatusFederationService;
+  private MeterRegistry meterRegistry;
 
   private PollingFederationService service;
 
   @Before
   public void setUp() {
     service = new PollingFederationService(populateStatusFederationService);
+    meterRegistry = new SimpleMeterRegistry();
+    service.setMeterRegistry(meterRegistry);
```
patduin, line 52: Shouldn't there be an actual test here?

patduin (approved):

Issue #264, opened 2023-03-02
Waggle-dance does not support DIGEST protocol


zzzzming95:
Hey, I have a problem. Our environment is based on kerberos. I backport this patch` https://github.com/ExpediaGroup/waggle-dance/commit/a4ea6950eb755ca279e405b3730a6f9a0a28f1dc ` to `3.10.10` .

Now we can using waggle-dance via kerberos protocol like hive-client or spark-client.

However, when I want to use hiveserver2 or trino based on the delegation-token mechanism, the connection fails.

The authentication mechanism of hiveserver2 is to use the kerberos protocol connection for the first time, obtain the delegation-token, and then add the token to the UGI, `ugi. addToken (delegationToken);` Later, the connection will be made through the 'DIGEST' protocol. 

The following is the code for the client to create different protocols.
```
        public TTransport createClientTransport(String principalConfig, String host, String methodStr, String tokenStrForm, final TTransport underlyingTransport, final Map<String, String> saslProps) throws IOException {
            final AuthMethod method = (AuthMethod)AuthMethod.valueOf(AuthMethod.class, methodStr);
            TTransport saslTransport = null;
            switch(method) {
            case DIGEST:
                Token<DelegationTokenIdentifier> t = new Token();
                t.decodeFromUrlString(tokenStrForm);
                saslTransport = new TSaslClientTransport(method.getMechanismName(), (String)null, (String)null, "default", saslProps, new HadoopThriftAuthBridge.Client.SaslClientCallbackHandler(t), underlyingTransport);
                return new TUGIAssumingTransport(saslTransport, UserGroupInformation.getCurrentUser());
            case KERBEROS:
                String serverPrincipal = SecurityUtil.getServerPrincipal(principalConfig, host);
```

In the server log of waggle-dance, there will be the following error.

```
java.lang.NullPointerException: null
    at org.apache.hadoop.security.SaslRpcServer.getIdentifier(SaslRpcServer.java:198) ~[hadoop-common-2.10.1.jar!/:?]
    at org.apache.hadoop.hive.thrift.HadoopThriftAuthBridge$Server$TUGIAssumingProcessor.process(HadoopThriftAuthBridge.java:573) ~[hive-shims-common-2.3.7.jar!/:2.3.7]
    at org.apache.thrift.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:286) [libthrift-0.9.3.jar!/:0.9.3]
[...]

PR #266, opened 2023-04-01
Support kerberos and delegation-token

zzzzming95:
### :pencil: Description

The authentication of our environment is all based on kerberos. I can see the kerberos patch in : https://github.com/ExpediaGroup/waggle-dance/commit/a4ea6950eb755ca279e405b3730a6f9a0a28f1dc

But the earlier patch for kerberos has some flaws, such it not support delegation-token auth , using the duplicate `HadoopThriftAuthBridge.Server`. We fix it and now it work well in our env.

### :link: Related Issues
https://github.com/ExpediaGroup/waggle-dance/issues/264


patduin:
> @abhimanyugupta07 @javsanbel2 cc

Thanks for the contribution! We'll try to have some eyes on this asap. 
Just a quick FYI we can run our tests but we don't have a Kerberos environment ourselves so those specific changes we'll have to rely on your tests.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -87,8 +99,26 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
           return Lists.newArrayList(user);
         }
+      case "get_delegation_token":
+        try {
+          base.open(cachedUgi);
+          Object token = doRealCall(method, args, attempt);
+          if (useSasl) {
+            this.delegationToken = (String) token;
+            base.close();
+            String newTokenSignature = base.getHiveConfValue(HiveConf.ConfVars.METASTORE_TOKEN_SIGNATURE.varname,
```
abhimanyugupta07, line 109: why do you have to set the default value "WAGGLEDANCETOKEN" here ?

are you not able to set it in the hive-site.xml and use it here ?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -132,6 +162,33 @@ private void reconnectIfDisconnected() {
...
+      UserGroupInformation currUser = UserGroupInformation.getCurrentUser();
+      if (delegationToken == null && currUser != UserGroupInformation.getLoginUser() && useSasl) {
+        String currShortName = currUser.getShortUserName();
+        Method getTokenMethod =
+                ThriftHiveMetastore.Iface.class.getMethod("get_delegation_token",
+                        String.class, String.class);
+        Object[] args = new Object[2];
+        args[0] = currShortName;
+        args[1] = currShortName;
+        String token = (String) doRealCall(getTokenMethod, args, 0);
```
patduin, line 177: This next few lines are duplicated above maybe we can refactor this in a method

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -132,6 +162,33 @@ private void reconnectIfDisconnected() {
...
+        base.close();
+        String newTokenSignature = base.getHiveConfValue(HiveConf.ConfVars.METASTORE_TOKEN_SIGNATURE.varname,
+                tokenSignature);
+        base.setHiveConfValue(HiveConf.ConfVars.METASTORE_TOKEN_SIGNATURE.varname,
+                newTokenSignature);
+        Utils.setTokenStr(currUser, token, newTokenSignature);
+        this.delegationToken = token;
+      }
+      if (!base.isOpen()) {
```
patduin, line 186: move this in inside the if above so we don't change existing logic.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -92,6 +92,12 @@ public PrefixBasedDatabaseMappingService(
 
   private void add(AbstractMetaStore metaStore) {
     MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(metaStore);
+    // connect metastore first
+    if (!metaStoreMapping.isAvailable()) {
```
patduin, line 96: why was this added?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/SaslHelper.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.hive.service.auth.HiveAuthFactory;
+import org.apache.hive.service.auth.PlainSaslHelper;
+import org.apache.hive.service.auth.SaslQOP;
+import org.apache.thrift.transport.TSaslServerTransport;
+import org.apache.thrift.transport.TTransportException;
+import org.apache.thrift.transport.TTransportFactory;
+
+public class SaslHelper {
+  public static HadoopThriftAuthBridge.Server createSaslServer(HiveConf conf) throws TTransportException {
+    HadoopThriftAuthBridge.Server saslServer = saslServer = null;
```
patduin, line 42:
```
HadoopThriftAuthBridge.Server saslServer = null;
```

patduin (changes requested):
Hi this looks good, just some small comments please have a look. Once merged we can get roll a release.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: Hmm not sure I follow, this method just adds a Metastore to the list of configured metastores and you're just logging a warning. 
You've already changed the code to do the delegate connection here:
[DefaultMetaStoreClientFactory.java](https://github.com/ExpediaGroup/waggle-dance/pull/266/files#diff-ee675b8d870d29e0855316cdace45011de4a0a97a308fe7cefbabb4937baa63a)
Which is the correct place to add the logic. 

Adding this here just forces a client.isOpen/getStatus call earlier in the process. Might not be needed...

patduin: The call `StaticDatabaseMappingService` is done because we have to get the database e.g. make a real connection.
Here I think we can defer till we get a use call (Which will do the getDelegate then) but it might not have a lot of impact. We can leave it an see.

patduin (approved):
@abhimanyugupta07 have a look please if ok, we can merge this.

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
+![Bee waggle-dancing on a hive.](logo.png "Federating Hive Meta Stores.")
+
+# Additional instruction to use Waggle Dance in kerberized environment
```
massdosage, line 3:
```suggestion
# Additional instructions to use Waggle Dance in a Kerberized environment
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+### Process
+
+In a kerberos environment, after a client requests waggle-dance, waggle-dance requests the proxy user's token from the metastore and then uses that token to communicate with the metastore.
```
massdosage, line 8:
```suggestion
In a Kerberos environment a client make a request to Waggle Dance which in turn requests the proxy user's token from the metastore and then uses this token to communicate with the metastore.
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+This is necessary in some scenarios that require permission authentication. Such as the `create_table` API that requires the proxy user to create hdfs directories.
```
massdosage, line 10:
```suggestion
This is necessary in certain scenarios that need authentication - for example the `create_table` API that requires the proxy user to create HDFS directories.
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+![image](https://user-images.githubusercontent.com/13965087/229339323-260f3c17-46c0-4471-81d2-cdbcfa0fe3ce.png)
```
massdosage, line 12: We should move a cleaned up version of this diagram to this repository instead of an external link.

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+In addition, because kerberos authentication requires a delegation-token to proxy as other users. The proxy user of the session is shared globally, which means we need to make all Hive Metastores share a set of delegation-token storage so that a single delegation-token can be authenticated by multiple Metastores.
```
massdosage, line 14:
```suggestion
In addition, because Kerberos authentication requires a delegation-token to proxy as other users, the proxy user of the session is shared globally. This means we need to make all Hive Metastores share a set of delegation-token storage so that a single delegation-token can be authenticated by multiple Metastores.
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+**One solution is to use zookeeper to store tokens for all Hive Metastores, which is essential.**
```
massdosage, line 16:
```suggestion
**One solution is to use Zookeeper to store tokens for all Hive Metastores**
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Prerequisites
+
+* Kerberized claster:
```
massdosage, line 20:
```suggestion
* Kerberized cluster:
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+  active KDC,
+  some required properties in configuration files of hadoop services
```
massdosage, line 22:
```suggestion
  some required properties in configuration files of Hadoop services
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+* User account with privileges in ipa
```
massdosage, line 23: What is `ipa`?

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+* Zookeeper to store delegation-token (Recommend)
```
massdosage, line 24:
```suggestion
* Zookeeper to store delegation-token (Recommended)
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Configuration
+
+Waggle Dance does not read hadoop's *core-site.xml* so a general property providing kerberos auth should be added to
+the Hive configuration file *hive-site.xml*:
```
massdosage, line 29:
```suggestion
Waggle Dance does not read Hadoop's `core-site.xml` so a general property providing Kerberos auth should be added to
the Hive configuration file `hive-site.xml`:
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+<property>
+  <name>hadoop.security.authentication</name>
+  <value>KERBEROS</value>
+</property>
+```
+
+
+Besides Waggle Dance needs a keytab file to communicate with the Metastore so following properties should be present:
```
massdosage, line 39:
```suggestion
Waggle Dance also needs a keytab file to communicate with the Metastore so the following properties should be present:
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+```
+<property>
+  <name>hive.metastore.sasl.enabled</name>
+  <value>true</value>
+</property>
+<property>
+  <name>hive.metastore.kerberos.principal</name>
+  <value>hive/_HOST@DEV.DF.SBRF.RU</value>
```
massdosage, line 47: Can we make this more generic?

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+</property>
+<property>
+  <name>hive.metastore.kerberos.keytab.file</name>
+  <value>/etc/hive.keytab</value>
+</property>
+```
+
+In addition, all metastores need to use the zk shared token:
```
massdosage, line 55:
```suggestion
In addition, all metastores need to use the Zookeeper shared token:
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+    <name>hive.cluster.delegation.token.store.zookeeper.connectString</name>
+    <value>zk1:2181,zk2:2181,zk3:2181</value>
+  </property>
+  <property>
+    <name>hive.cluster.delegation.token.store.zookeeper.znode</name>
+    <value>/hive/token</value>
+  </property>
+```
+
+If you are intending to use a beeline client, following properties may be valuable:
```
massdosage, line 71:
```suggestion
If you are intending to use a Beeline client, the following properties may be valuable:
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+  <name>hive.server2.transport.mode</name>
+  <value>http</value>
+</property>
+<property>
+  <name>hive.server2.authentication</name>
+  <value>KERBEROS</value>
+</property>
+<property>
+  <name>hive.server2.authentication.kerberos.principal</name>
+  <value>hive/_HOST@DEV.DF.SBRF.RU</value>
```
massdosage, line 83: Same as above

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+<property>
+  <name>hive.server2.enable.doAs</name>
+  <value>false</value>
+</property>
+```
+
+
+### Running
+
+The Waggle Dance should be starting as privileged user with a fresh keytab.
```
massdosage, line 98:
```suggestion
Waggle Dance should be started by a privileged user with a fresh keytab.
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+If Waggle Dance throws a GSS exception, you have problem with the keytab file.
+Try to perform `kdestroy` and `kinit` operations and check for a keytab file ownership.
```
massdosage, line 101:
```suggestion
Try to perform `kdestroy` and `kinit` operations and check the keytab file ownership flags.
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+
+If the Metastore throws an exception with code -127, Waggle Dance uses wrong authentication policy.
```
massdosage, line 103:
```suggestion
If the Metastore throws an exception with code -127, Waggle Dance is probably using the wrong authentication policy.
```

HowToKerberize.md
```diff
@@ -0,0 +1,106 @@
...
+Check hive-conf.xml for a correct configuration and make sure that HIVE_HOME and HIVE_CONF_DIR are defined.
```
massdosage, line 104:
```suggestion
Check the values in `hive-conf.xml` and make sure that HIVE_HOME and HIVE_CONF_DIR are defined.
```

waggle-dance-core/pom.xml
```diff
@@ -350,6 +350,21 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
massdosage, line 353: This section is for `test` dependencies. Is this only needed for tests? If so `<scope>test</scope>` should be added. If not it should be moved to the `hive` section between lines 118 and 163.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/SaslHelper.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.hadoop.hive.thrift.HadoopThriftAuthBridge;
+import org.apache.hadoop.hive.thrift.HiveDelegationTokenManager;
+import org.apache.hive.service.auth.HiveAuthFactory;
+import org.apache.hive.service.auth.PlainSaslHelper;
+import org.apache.hive.service.auth.SaslQOP;
+import org.apache.thrift.transport.TSaslServerTransport;
+import org.apache.thrift.transport.TTransportException;
+import org.apache.thrift.transport.TTransportFactory;
+
+public class SaslHelper {
```
massdosage, line 40: Since this is a "util" class it should be made final and have a private no-args constructor added to enforce only static usage.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/util/SaslHelper.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public static Map<String, String> getSaslProperties(HiveConf hiveConf) {
+    Map<String, String> saslProps = new HashMap<String, String>();
+    SaslQOP saslQOP = SaslQOP.fromString(hiveConf.getVar(HiveConf.ConfVars.HIVE_SERVER2_THRIFT_SASL_QOP));
+    saslProps.put(Sasl.QOP, saslQOP.toString());
+    saslProps.put(Sasl.SERVER_AUTH, "true");
+    return saslProps;
+  }
+}
```
massdosage, line 117: Please add a closing line break.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -132,6 +158,36 @@ private void reconnectIfDisconnected() {
...
+        this.delegationToken = token;
+        if (!base.isOpen()) {
+          base.open(cachedUgi);
+        }
+      }
+      return delegationToken;
+    }
+
+    private void setTokenStr2Ugi(UserGroupInformation currUser, String token) throws IOException {
+      String newTokenSignature = base.getHiveConfValue(HiveConf.ConfVars.METASTORE_TOKEN_SIGNATURE.varname,
```
massdosage, line 185: I'd suggest replacing the `getHiveConfValue` and `setHiveConfValue` methods could with a new method, something like `generateNewTokenSignature` which does all of this?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -52,6 +60,10 @@ private ReconnectingMetastoreClientInvocationHandler(
       this.name = name;
       this.maxRetries = maxRetries;
       this.base = base;
+      if (base != null)
+        this.useSasl = base.getHiveConfBool(HiveConf.ConfVars.METASTORE_USE_THRIFT_SASL);
```
massdosage, line 64: We could rename this method to `isSaslEnabled`, remove the argument and then we won't need to expose the fact that the conf is being used etc.

massdosage (changes requested):

patduin:
> @abhimanyugupta07 have a look please if ok, we can merge this.

or @massdosage  :) thanks for reviewing

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
abhimanyugupta07: can you elaborate a bit more on why do you have to mask a default value for this property ?

abhimanyugupta07: If you put a null/empty check and only set it when it is actually set in the hive-site.xml, will it not be sufficient ?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: Hi been testing with this call and please remove this. It will impact normal calls.
Considers federating to 10+ metastores we'll be doing this isAvailable check for all of them and if any of those are slow to respond we've added unnecessary latency. These calls should happen lazily only if needed. 
Which I think will work ok because of your changes in the `DefaultMetaStoreClientFactory`.

patduin (changes requested):
Did some checks please have a look :)

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
patduin: This part of the code should not have any knowledge of SASL or non SASL connection it should treat it all the same.

If this call is needed for Kerberos then something might need to be added in: `TProcessorFactorySaslDecorator`
Specifically I am thinking that perhaps you can inject: `NotifyingFederationService notifyingFederationService` and ` MetaStoreMappingFactory metaStoreMappingFactory` in the `MetaStoreProxyServer`

Then add it to call here: https://github.com/ExpediaGroup/waggle-dance/pull/266/files#diff-b9afe479963c18cd3af539fa88f9b7b728510cf1ad2b24a2ee26658a423770c1R210

In the `TProcessorFactorySaslDecorator.getProcess()`
Do a loop over notifyingFederationService.getAll() for each create a mapping and check availability. That way the extra work is only done for kerberos and won't affect anything else.

HowToKerberize.md
patduin: I think what @massdosage means is to upload the image as part of the PR and reference it just like we do with: https://github.com/ExpediaGroup/waggle-dance/blob/main/system-diagram.png

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -92,10 +93,16 @@ public PrefixBasedDatabaseMappingService(
 
   private void add(AbstractMetaStore metaStore) {
     MetaStoreMapping metaStoreMapping = metaStoreMappingFactory.newInstance(metaStore);
+
     DatabaseMapping databaseMapping = createDatabaseMapping(metaStoreMapping);
 
     if (metaStore.getFederationType() == PRIMARY) {
       primaryDatabaseMapping = databaseMapping;
+      if (!metaStoreMapping.isAvailable()) {
```
patduin, line 101: ok I still think this is brittle because the code relies on a side-effect of `isAvailable` but ok if this is a simple solution for kerberizing.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -430,6 +437,7 @@ public void close() throws IOException {
       for (MetaStoreMapping metaStoreMapping : mappingsByPrefix.values()) {
         metaStoreMapping.close();
       }
+      TokenWrappingHMSHandler.removeToken();
```
patduin, line 440: this line should go: it's untestable and there is lifecycle where this is calls via the `TTransportMonitor` which call closes on the TokenWrappingHMSHandler which should remove the token already.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -408,6 +409,7 @@ public void close() throws IOException {
       for (MetaStoreMapping metaStoreMapping : mappingsByMetaStoreName.values()) {
         metaStoreMapping.close();
       }
+      TokenWrappingHMSHandler.removeToken();
```
patduin, line 412: please remove this calls should no know anything about tokens

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TSetIpAddressProcessorFactory.java
```diff
@@ -60,8 +60,11 @@ public TProcessor getProcessor(TTransport transport) {
       CloseableIHMSHandler baseHandler = federatedHMSHandlerFactory.create();
       IHMSHandler handler = newRetryingHMSHandler(ExceptionWrappingHMSHandler.newProxyInstance(baseHandler), hiveConf,
           false);
+      boolean useSASL = hiveConf.getBoolVar(HiveConf.ConfVars.METASTORE_USE_THRIFT_SASL);
+      IHMSHandler tokenHandler = newRetryingHMSHandler(TokenWrappingHMSHandler.newProxyInstance(handler, useSASL), hiveConf,
```
patduin, line 64: wrap this line in: 
```
if (useSasl) {
IHMSHandler tokenHandler = newRetryingHMSHandler(TokenWrappingHMSHandler.newProxyInstance(handler), hiveConf, false);
transportMonitor.monitor(transport, tokenHandler);
return new TSetIpAddressProcessor<>(tokenHandler);
} else {
 transportMonitor.monitor(transport, baseHandler);
 return new TSetIpAddressProcessor<>(handler);
}

File: waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java [hunk 35]
```diff
@@ -87,7 +100,22 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
+            base.close();
+            setTokenStr2Ugi(UserGroupInformation.getCurrentUser(), (String) token);
+            base.open(cachedUgi);
+          }
+          return token;
+        } catch (IOException | TException e) {
+          throw new MetastoreUnavailableException("Couldn't setup delegation token in the ugi: ", e);
+        }
       default:
+        genToken();
```
patduin, line 118: shouldn't this only be done when useSasl=true?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TokenWrappingHMSHandler.java
```diff
@@ -0,0 +1,93 @@
...
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
+import org.apache.hadoop.hive.metastore.IHMSHandler;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TokenWrappingHMSHandler implements InvocationHandler {
```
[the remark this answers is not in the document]
patduin: Would be nice if it had a unit test :)

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactoryTest.java
```diff
@@ -107,7 +108,7 @@ public void defaultMethodCallThrowsTransportExceptionRetries() throws TException
 
     String result = iface.getName();
     assertThat(result, is("ourName"));
-    verify(base).open(null);
+    verify(base, Mockito.times(2)).open(null);
```
patduin, line 111: I think is back to 1 call???

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -132,6 +160,22 @@ private void reconnectIfDisconnected() {
       }
     }
 
+    private void genToken() throws Throwable {
```
patduin, line 163: I quite like the approach of the TokenWrappingHMSHandler being separate for this code the client I suggest we do the same and create a `SaslMetastoreClientHander implements  InvocationHandler`  
This class newInstance methos can check if useSasl=true and wrap the `ReconnectingMetastoreClientInvocationHandler` inside the Sasl handler. That way we don't affect the non sasl code paths at all.

Let me know what you think.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -155,9 +226,17 @@ CloseableThriftHiveMetastoreIface newInstance(
...
       ThriftMetastoreClientManager base) {
     ReconnectingMetastoreClientInvocationHandler reconnectingHandler = new ReconnectingMetastoreClientInvocationHandler(
-        name, reconnectionRetries, base);
-    return (CloseableThriftHiveMetastoreIface) Proxy
-        .newProxyInstance(getClass().getClassLoader(), INTERFACES, reconnectingHandler);
+            name, reconnectionRetries, base);
+    if (base.isSaslEnabled()) {
+      CloseableThriftHiveMetastoreIface ifaceReconnectingHandler = (CloseableThriftHiveMetastoreIface) Proxy
+              .newProxyInstance(getClass().getClassLoader(), INTERFACES, reconnectingHandler);
+      // Warpping the SaslMetastoreClientHander to handle delegation token if using sasl
```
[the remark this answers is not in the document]
patduin: Warpping=wrapping

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactory.java
```diff
@@ -155,9 +226,17 @@ CloseableThriftHiveMetastoreIface newInstance(
       int reconnectionRetries,
       ThriftMetastoreClientManager base) {
     ReconnectingMetastoreClientInvocationHandler reconnectingHandler = new ReconnectingMetastoreClientInvocationHandler(
-        name, reconnectionRetries, base);
-    return (CloseableThriftHiveMetastoreIface) Proxy
-        .newProxyInstance(getClass().getClassLoader(), INTERFACES, reconnectingHandler);
+            name, reconnectionRetries, base);
+    if (base.isSaslEnabled()) {
```
patduin, line 230: we have a test `DefaultMetastoreClientFactory.closeNullBase` which test that base == null, this test I think can be deleted we don't check for null anywhere except in the close method.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/client/DefaultMetaStoreClientFactoryTest.java
```diff
@@ -122,7 +123,7 @@ public void set_ugi_before_call() throws Exception {
     String name = iface.getName();
 
     assertThat(name, is("ourName"));
-    verify(base).open(TEST_ARGS);
+    verify(base, Mockito.times(2)).open(TEST_ARGS);
```
patduin, line 126: this is called once

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/TokenWrappingHMSHandler.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.waggledance.server;
+
+import java.lang.reflect.*;
```
patduin, line 18: organize imports please

patduin (approved):

patduin:
Thank you @zzzzming95 This looks good!
 I think @massdosage still has some open comments, please have a look.
@massdosage @abhimanyugupta07 if you both are happy please approve?

massdosage:
LGTM, thanks for the contribution @zzzzming95

abhimanyugupta07 (approved):

PR #270, opened 2023-05-11
Remove bean overriding

RickArt:
### :pencil: Description
- Removing bean overriding as a default in favor of keeping a cleaner bean definition.
- Flagged v3.10.12 as Yanked as it was wrongly deployed by mistake from a feature branch.
- Flagged v3.10.13 as Yanked as it was changed by mistake from a maven release action, no deployment to maven central was made. optionally we could revert the commits from the action since it was not actually deployed (?)

### :link: Related Issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-## [3.10.12] - 2023-05-04
+## [3.10.14] - 2023-05-11
+### Changed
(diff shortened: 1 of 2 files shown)


patduin (approved):

patduin:
> * Flagged v3.10.13 as Yanked as it was changed by mistake from a maven release action, no deployment to maven central was made. optionally we could revert the commits from the action since it was not actually deployed (?)

yeah we can discuss, it's also not a big deal to just release the next version.

PR #275, opened 2023-05-22
Add custom annotation to track method execution

prakhar10:
### :link: Related Issues


waggle-dance-core/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
patduin, line 1: there should not be any log4j config file here please remove, this is created by the packaging module: https://github.com/ExpediaGroup/waggle-dance/blob/main/waggle-dance/src/main/resources/log4j2.xml

patduin (changes requested):
This will generate way too many log messages with most of them not being very informative.

patduin:
Talked offline, gonna close this till we have a better understanding.

PR #293, opened 2023-09-10
Supplement and modify some methods in Hive3.x

weitens:
### Description:
PR is a proposal to adapt to Hive3. I have maintained the majority of the methods in the original FederatedHMSHandler and made temporary implementations of the missing methods in this class (which may not be as appropriate), currently it can support compilation and can support some basic syntax of hive normally
In addition, for the db returned by hive3, if the returned item carries catalog information, I will extract db_name
Welcome any feedback, especially some suggestions about routing and forwarding logic (I'm not sure this makes sense for everyone)


patduin:
Thanks for contributing @weitens there are some conflicts in this PR could you please fix them. We'll try to have a look at this in coming weeks.

weitens:
> Thanks for contributing @weitens there are some conflicts in this PR could you please fix them. We'll try to have a look at this in coming weeks.

Thank you for your reply. I will fix these conflicts as soon as possible.

PR #295, opened 2023-09-26
Fix  primary database not filter

flaming-archer:
### :pencil: Description
When database-resolution set manual，primary database also config some database,so to determine whether to add the fed database, it should be necessary to filter out the ones that are not in the primary database.

### :link: Related Issues

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
@@ -85,6 +85,11 @@ public class StaticDatabaseMappingService implements MappingEventListener {
       QueryMapping queryMapping) {
     this.metaStoreMappingFactory = metaStoreMappingFactory;
     this.queryMapping = queryMapping;
+    mappingsByMetaStoreName = Collections.synchronizedMap(new LinkedHashMap<>());
+    mappingsByDatabaseName = Collections.synchronizedMap(new LinkedHashMap<>());
+    databaseMappingToDatabaseList = new ConcurrentHashMap<>();
+    databaseToTableAllowList = new ConcurrentHashMap<>();
+
     primaryDatabasesCache = CacheBuilder
         .newBuilder()
         .expireAfterAccess(1, TimeUnit.MINUTES)
@@ -94,17 +99,14 @@ public class StaticDatabaseMappingService implements MappingEventListener {
           @Override
           public List<String> load(String key) throws Exception {
             if (primaryDatabaseMapping != null) {
-              return primaryDatabaseMapping.getClient().get_all_databases();
+              return new StaticDatabaseMappingPanopticOperationHandler()
+                  .getPrimaryAllDatabases();
             } else {
               return Lists.newArrayList();
             }
           }
         });
 
-    mappingsByMetaStoreName = Collections.synchronizedMap(new LinkedHashMap<>());
-    mappingsByDatabaseName = Collections.synchronizedMap(new LinkedHashMap<>());
-    databaseMappingToDatabaseList = new ConcurrentHashMap<>();
-    databaseToTableAllowList = new ConcurrentHashMap<>();
     for (AbstractMetaStore federatedMetaStore : initialMetastores) {
       add(federatedMetaStore);
     }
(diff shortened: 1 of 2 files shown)


flaming-archer:
@patduin Could you help to take a look at this? It hasn't really changed anything. Let the primaly database be used as a filter, only for its configuration, not for all.

patduin (changes requested):
Please add some tests. Perhaps consider jUnit and WaggeDanceIntegrationTest for this usecase.

flaming-archer:
> Please add some tests. Perhaps consider jUnit and WaggeDanceIntegrationTest for this usecase.

Sorry for late replay, I will add some tests later.

flaming-archer:
@patduin hi,I  have been very busy lately and finally have time to write. I have added a test, please help me review it.

patduin:
Thanks! So have I! No worries thanks for adding I'll try to find some time
coming week to have a look.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
```diff
@@ -94,17 +100,14 @@ public StaticDatabaseMappingService(
           @Override
           public List<String> load(String key) throws Exception {
             if (primaryDatabaseMapping != null) {
-              return primaryDatabaseMapping.getClient().get_all_databases();
+              return new StaticDatabaseMappingPanopticOperationHandler()
```
patduin, line 103: Is this the issue? We don't filter here?

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingServiceTest.java
```diff
@@ -458,6 +459,25 @@ public void panopticOperationsHandlerGetAllDatabases() {
...
+
+    primaryMetastore.setMappedDatabases(Lists.newArrayList("primary_db1"));
+
+    service = new StaticDatabaseMappingService(metaStoreMappingFactory,
+        Arrays.asList(primaryMetastore, federatedMetastore), queryMapping);
+
+
+    StaticDatabaseMappingPanopticOperationHandler handler =
+        (StaticDatabaseMappingPanopticOperationHandler) service.getPanopticOperationHandler();
+    assertThat(handler.getPrimaryAllDatabases(), is(Lists.newArrayList("primary_db1")));
```
patduin, line 478: This asserts that the code you added work, but it doesn't explain to me what the behaviour is that you want to change.

Please write a test that tests the difference in behaviour you are after?

It might help if you explain which e.g. hive queries  or HMS methods you're trying to call that behave incorrectly?

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java
flaming-archer: yes, we don't filter. I just did a filtering job.

waggle-dance-core/src/test/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingServiceTest.java
flaming-archer: I have write a test now, pls review .

> This asserts that the code you added work, but it doesn't explain to me what the behaviour is that you want to change.
> 
> Please write a test that tests the difference in behaviour that you like to implement?
> 
> It might help if you explain which e.g. hive queries or HMS methods you're trying to call that behave incorrectly?

flaming-archer:
@patduin Could you take a look at this. If ok, please merge it.

patduin:
> flaming-archer

Hi,

I've been busy on this as it is a bit tricky.

Thanks for writing the test that helped me understand what you were trying to fix!

Looking into this I thought the use of this panoptic class was a bit weird as we do filtering already when we add the metastore to the list ([using the Allowlist](https://github.com/ExpediaGroup/waggle-dance/blob/hive-3.x/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/StaticDatabaseMappingService.java#L120)).
I've made a different solution which also removes this cache which I believe is not necessary anymore. 
I think it's an historical piece where we had a usecase that created databases on the metastore directly (outside of WD) and we wanted to pick those up, before we had mapping implemented.
The StaticDatabaseMappingService gets created and loaded for every request so I don't think this cache solves anything that a user can't easily solve by just reconnecting.

Why is this all tricky?
When doing static (manual) mapping things become hard when new databases are created that cause a crash, if we just crashloop WD throwing exception that the mapping should be updated. This is why I prefer the prefixes less risk of this.
I've [raised a PR](https://github.com/ExpediaGroup/waggle-dance/pull/302) for this here, please have a look. And big disclaimer (sadly) I think I mentioned this before we don't run Hive3 at our company so we do not have QA setup for this branch just the unit test, please run your own QA before accepting this.
As always thanks for the contribution keen to hear your thoughts.

patduin:
Closed in favour of: https://github.com/ExpediaGroup/waggle-dance/pull/302

PR #297, opened 2023-09-29
Modify the processing of some variables in hive-3.x

weitens:
### :pencil: Description
PR is a proposal to adapt to Hive3. It mainly includes **two** commits, which are described as follows:
1. When I used the waggle-dance to form a hive3 federation in MANUAL mode, I found that using **SHOW DATABASES** would repeatedly display the **default** based on the number of Hive Metastore. I don't know if it was caused by an error in my configuration file. I have performed a deduplication operation to optimize the display effect.
2. For the database name returned by hive3, if the returned item carries catalog information, it will not be found in the database map of waggle-dance and will trigger an exception that does not exist in the database. Therefore, I imitated the processing method of hive3 and extracted **database name** from the passed database values.
Welcome any feedback, I believe we can make waggle-dance better

### DIFF
--- a/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
+++ b/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
@@ -26,6 +26,7 @@ import org.apache.hadoop.hive.metastore.MetaStoreEventListener;
 import org.apache.hadoop.hive.metastore.RawStore;
 import org.apache.hadoop.hive.metastore.TransactionalMetaStoreEventListener;
 import org.apache.hadoop.hive.metastore.Warehouse;
+import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.apache.hadoop.hive.metastore.api.AbortTxnRequest;
 import org.apache.hadoop.hive.metastore.api.AbortTxnsRequest;
 import org.apache.hadoop.hive.metastore.api.AddCheckConstraintRequest;
@@ -362,6 +363,7 @@ class FederatedHMSHandler extends FacebookBase implements CloseableIHMSHandler {
   @Override
   @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
   public Database get_database(String name) throws NoSuchObjectException, MetaException, TException {
+    name = MetaStoreUtils.parseDbName(name, conf)[1];
     log.info("Fetching database {}", name);
     DatabaseMapping mapping = databaseMappingService.databaseMapping(name);
     log.info("Mapping is '{}'", mapping.getDatabasePrefix());
@@ -373,6 +375,7 @@ class FederatedHMSHandler extends FacebookBase implements CloseableIHMSHandler {
   @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
   public void drop_database(String name, boolean deleteData, boolean cascade)
       throws NoSuchObjectException, InvalidOperationException, MetaException, TException {
+    name = MetaStoreUtils.parseDbName(name, conf)[1];
     DatabaseMapping mapping = checkWritePermissions(name);
     mapping.getClient().drop_database(mapping.transformInboundDatabaseName(name), deleteData, cascade);
   }
@@ -488,6 +491,7 @@ class FederatedHMSHandler extends FacebookBase implements CloseableIHMSHandler {
       boolean deleteData,
       EnvironmentContext environment_context)
       throws NoSuchObjectException, MetaException, TException {
+    dbname = MetaStoreUtils.parseDbName(dbname, conf)[1];
     DatabaseMapping mapping = checkWritePermissionsAndCheckTableAllowed(dbname, name);
     mapping
         .getClient()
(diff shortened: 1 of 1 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
```diff
@@ -476,6 +479,7 @@ public void create_table_with_constraints(Table tbl, List<SQLPrimaryKey> primary
   @Loggable(value = Loggable.DEBUG, skipResult = true, name = INVOCATION_LOG_NAME)
   public void drop_table(String dbname, String name, boolean deleteData)
       throws NoSuchObjectException, MetaException, TException {
+    name = MetaStoreUtils.parseDbName(name, conf)[1];
```
patduin, line 482: I think added testcases will help getting understanding what you're trying to fix.

patduin (changes requested):
I've added some comments

weitens:
> 1. Could you please share you configuration because this should not happen (maybe a hive 3 thing) but that's a bit strange.Not sure if filtering on our side is the way forward here.
> 2. Do you mean something like hive.my_db_name.my_table causes issues?

My profile content is as follows:
```
primary-meta-store:
  access-control-type: READ_AND_WRITE_AND_CREATE
  latency: 0
  mapped-databases:
  - default
  - wd-test1
  name: local
  remote-meta-store-uris: thrift://hadoop101:9083
federated-meta-stores:
- access-control-type: READ_AND_WRITE_ON_DATABASE_WHITELIST
  database-prefix: remote1
  latency: 0
  mapped-databases:
  - wd-test2
  name: remote1
  remote-meta-store-uris: thrift://hadoop102:9083
  writable-database-white-list:
  - wd-test2
- access-control-type: READ_AND_WRITE_ON_DATABASE_WHITELIST
  database-prefix: default2
  latency: 0
  mapped-databases:
  - hadoop101
  name: wd-test3
  remote-meta-store-uris: thrift://hadoop103:9083
  writable-database-white-list:
  - wd-test3
- access-control-type: READ_AND_WRITE_ON_DATABASE_WHITELIST
  database-prefix: default3
  latency: 0
  mapped-databases:
  - wd-test4
  name: remote3
  remote-meta-store-uris: thrift://hadoop104:9083
  writable-database-white-list:
  - wd-test4

```

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
weitens: I am truly sorry, this should be dbname instead of name. I will modify the file content as soon as possible

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
```diff
@@ -79,6 +80,7 @@ protected List<String> getAllDatabases(
     List<String> result = getPanopticOperationExecutor()
         .executeRequests(allRequests, GET_DATABASES_TIMEOUT, "Can't fetch databases by pattern: {}");
     log.info("All Databases Result={}", result);
+    result = result.stream().distinct().collect(Collectors.toList());
```
patduin, line 83: Please remove this and write a test to reproduce this erorr. https://github.com/ExpediaGroup/waggle-dance/blob/hive-3.x/waggle-dance-integration-tests/src/test/java/com/hotels/bdp/waggledance/WaggleDanceIntegrationTest.java

Best to raise a separate PR for this issue

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
patduin: I meant in general I'm not sure what sure what this code does without reading it:  `MetaStoreUtils.parseDbName`
It will help if you write unit test for the changes you made so it's easier to understand why you made them. Thanks!

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/PanopticOperationHandler.java
weitens: Thank you for your suggestion. I have removed the commit.
I will try to reproduce the issue and raise a separate PR for this issue within the next week.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/server/FederatedHMSHandler.java
weitens: I will add corresponding unit tests for these modifications as soon as possible to improve their readability.

weitens: I'm sorry, for some reason, I confused the latest code generated tarball with the previous tarball, which led me to believe that the latest commit did not solve this problem.
The above issue has been resolved in a345e74e0a5120bbff03ac0fd90cab4d4ee74809. For the case where **@hive#default** carries catalog information, it uses `GrammarUtils.removeCatName(databaseName)` method to remove the catalog information. 
Thank you for your enthusiastic reply and suggestions. In the future, I will be more cautious with every PR I receive.

PR #300, opened 2023-10-20
Added timeout on isAvailable calls

patduin:
This PR helps in a situation where you have multiple metastores configured and connectivity is not always stable or metastore are not always responding.


mroark1m (approved):
LGTM

githubjianli (approved):

PR #308, opened 2024-01-23
Added extra option to add readOnly thrift HMS uri 

patduin:
**tldr;** Split traffic based on called HMS API method, e.g. `getTable` will go to a readOnly HMS and `alterTable` will go to readWrite HMS

The problem addressed here is running WD at scale. Generally our company deploys Waggle Dance as part of an Apiary Data lake: https://github.com/ExpediaGroup/apiary-data-lake.
This involves deploying ReadOnly and ReadWrite Metastores (HMS). 
For the primary (local) metastore waggle dance is configured to the ReadWrite instance which connects to a ReadWrite RDS backend. This means all traffic both read and writes end up on our ReadWrite RDS instance. This PR tries to split that traffic and move read traffic to ReadOnly instance. 
The benefit would be:
* easily scale ReadOnly RDS instances to handle more load
* Automatically redirect the traffic without user configuration changes. Lots of ETL do read and writes as part of their workflow it has proven difficult for users to fully switch to a ReadOnly instance only this PR makes the decision for them.


waggle-dance-api/src/main/java/com/hotels/bdp/waggledance/api/model/AbstractMetaStore.java
```diff
@@ -57,6 +57,7 @@ public abstract class AbstractMetaStore {
   private long latency = 0;
   private transient @JsonIgnore HashBiMap<String, String> databaseNameBiMapping = HashBiMap.create();
   private GlueConfig glueConfig;
+  private String readOnlyRemoteMetaStoreUris;
```
patduin, line 60: TODO add to ReadMe.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SplitTrafficMetastoreClientFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
+/**
+ * This class splits the traffic for read only calls (get* for instance getTable, getPartition) to the readOnly client
+ * and everything else will go to readWrite client.
+ */
+public class SplitTrafficMetastoreClientFactory {
```
patduin, line 27: TODO add test

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SplitTrafficMetastoreClientFactory.java
```diff
@@ -0,0 +1,87 @@
...
+        CloseableThriftHiveMetastoreIface readWrite,
+        CloseableThriftHiveMetastoreIface readOnly) {
+      this.readWrite = readWrite;
+      this.readOnly = readOnly;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+      switch (method.getName()) {
+      case "isOpen":
```
jmnunezizu, line 46: Nitpick: I'm always wary about using `switch` statements like this. Failing to `return` might cause problems.

README.md
```diff
@@ -167,6 +167,7 @@ The table below describes all the available configuration values for Waggle Danc
...
 | `primary-meta-store.database-name-mapping`              | No       | BiDirectional Map of database names and mapped name, where key=`<database name as known in the primary metastore>` and value=`<name that should be shown to a client>`. See the [Database Name Mapping](#database-name-mapping) section.|
 | `primary-meta-store.glue-config`                       | No       | Can be used instead of `remote-meta-store-uris` to federate to an AWS Glue Catalog ([AWS Glue](https://docs.aws.amazon.com/glue/index.html). See the [Federate to AWS Glue Catalog](#federate-to-aws-glue-catalog) section.|
+| `primary-meta-store.read-only-remote-meta-store-uris`    | No       | Can be used to configure an extra read-only endpoint for the primary Metastore. This is an optimization if your environment runs separate Metastore endpoints and traffic needs to be differted efficiently. Waggle Dance will direct traffic to the read-write or read-only endpoints based on the call being done. For instance `get_table` will be a read-only call but `alter_table` will be forwarded to the read-write Metastore.|
```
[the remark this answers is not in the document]
patduin: diverted 🤦

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SplitTrafficMetastoreClientFactory.java
patduin: yeah I somewhat agree, but we have a couple of invocation handlers doing the same thing so it will be easier to understand like this.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SplitTrafficMetastoreClientFactory.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        return null;
+      case "set_ugi":
+        Object result = doRealCall(readWrite, method, args);
+        // we skip the result for readOnly (it should always be the same).
+        doRealCall(readOnly, method, args);
+        return result;
+      default:
+        if (method.getName().startsWith("get")) {
+          return doRealCall(readOnly, method, args);
```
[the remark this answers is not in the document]
patduin: yeah we can, I'll add some logs if they get too chatty we might have to tune that.
I checked our metrics so we just plot tag the method call with the metastore name so we don't have the info there plus our invocation log also doesn't show this. So it's a good idea.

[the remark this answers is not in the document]
jmnunezizu: Excellent! Thanks.

README.md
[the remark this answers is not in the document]
jmnunezizu: You gave us a mashup of both! Haha.

waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/client/SplitTrafficMetastoreClientFactory.java
jmnunezizu: Yes, fair enough,

jmnunezizu (approved):

githubjianli (approved):

mroark1m (approved):

PR #331, opened 2024-10-03
replaced isAvailable call

patduin:
### :link: Related Issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
+## [3.13.4] - 2024-10-03
+### Fix
+- Removed call to primary and made it only when Kerberos (SASL) is enabled. 
+- Fixed Handler creation for SASL.
+
 ## [3.13.3] - 2024-10-02
 ### Added
-- Metric for monitoring open transports. `<prefix>.open_transports_gauge`
+- Metric for monitoring open transports. `<prefix>.com_hotels_bdp_waggledance_open_transports_gauge`
 ### Changed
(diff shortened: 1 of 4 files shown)


waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/mapping/service/impl/PrefixBasedDatabaseMappingService.java
```diff
@@ -97,11 +97,6 @@ private void add(AbstractMetaStore metaStore) {
 
     if (metaStore.getFederationType() == PRIMARY) {
       primaryDatabaseMapping = databaseMapping;
-      if (!metaStoreMapping.isAvailable()) {
```
patduin, line 100: This was added as part of Kerberos support: https://github.com/ExpediaGroup/waggle-dance/pull/266/files#diff-8c2a82b373b4f06981a5086c6e6f2cbafea8480451366fcf1ba3c1490450388a

For non Kerberos this is not needed and unwanted as Healtcheck connections just opening the processor will cause this to call the primary metastore. In production environments this is slowing things down. I moved it to the SASL code path.

mollonado (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
+## [3.13.4] - 2024-10-03
```
JayGreeeen, line 1:
```suggestion
## [3.13.4] - 2024-10-04
```

JayGreeeen (approved):

javsanbel2 (approved):

PR #336, opened 2025-02-11
Resolved the issue where the custom database prefix was lost after th…

l-jelly:
### :pencil: Description
Configure the primary-meta-store.database-prefix: usercluster2_ for the primary database. Connect to waggle-dance through hs2 and execute create database xxx. The primary database prefix is ​​lost when showing databases.
image

### :link: Related Issues


patduin (approved):
Great thank you!

abhimanyugupta07 (approved):

PR #344, opened 2025-05-23
fix: Function names with parameters cannot have a prefix.

KeeProMise:
### :pencil: Description
When a function has parameters, using prefix routing can prevent these functions from being prefixed, resulting in them not being found.

You can use the original code to run the newly added unit tests, and the bug can be reproduced.

### :link: Related Issues


KeeProMise:
@patduin @jmnunezizu Hi, if you have no more comments or questions, please help merge this PR, thank!

PR #358, opened 2026-06-25
Fix/glue lakeformation access denied get database

rpoluri:
### :pencil: Description
Handle LakeFormation Access Denied Excpetion when describing invalid Database and Tables in Glue Catalogs.

### :link: Related Issues


javsanbel2 (changes requested):
I think this should not be the place where this exception is handled

Have you considered handling the exception here? https://github.com/ExpediaGroup/aws-glue-data-catalog-client-for-apache-hive-metastore/tree/branch/waggle-dance

This library is used within Waggledance to connect to Glue as a thrift client

javsanbel2:
Done as part of https://github.com/ExpediaGroup/waggle-dance/pull/359

